PR #3, opened 2018-03-27
EVA-1112 Accession repository structure

jmmut:


eva-accession-ws/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+            <groupId>uk.ac.ebi.eva</groupId>
+            <artifactId>eva-accession-core</artifactId>
+            <version>1.0-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>uk.ac.ebi.eva</groupId>
+            <artifactId>eva-accession-core</artifactId>
+            <version>1.0-SNAPSHOT</version>
+            <scope>test</scope>
+            <type>test-jar</type>
```
cyenyxe: Is this "repeated" dependency really necessary?

eva-accession-ws/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+        </dependency>
+    </dependencies>
+
+    <repositories>
```
cyenyxe: Would it be enough to have these just in the parent POM?

eva-accession-ws/pom.xml
jmmut: well, if you don't put this, eva-accession-ws fails to compile, so yes. it's bringing the test classes, so it's not really the same

PR #4, opened 2018-03-27
EVA-1115 Implement ss model

jmmut:
ss is identified by:
-    Reference assembly
-    Taxonomy (for the sake of retrocompatibility, because dbSNP don't group submissions in a database based on the reference assembly)
-    Batch
-    Chromosome or contig
-    Start position
-    Reference allele
-    Alternate allele

changes:
-    getChromosome renamed as getContig (because the region could be either a chromosome or a contig)
-    Variant type replaced by reference and alternate alleles
-    VariantXXX renamed as SubmittedVariantXXX
-    add non-identifying field `supportedByEvidence`


eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariantModelSummaryFunction.java
```diff
@@ -17,16 +17,20 @@
...
 
-public class VariantModelSummaryFunction implements java.util.function.Function<VariantModel, String> {
+import java.util.function.Function;
+
+public class SubmittedVariantModelSummaryFunction implements Function<SubmittedVariantModel, String> {
 
     @Override
-    public String apply(VariantModel model) {
+    public String apply(SubmittedVariantModel model) {
         return new StringBuilder()
```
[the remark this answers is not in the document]
jmmut: adding a separator doesn't prevent this, actually. e.g.:
- project: "p_1", contig: "2"
- project: "p", contig: "1_2"

both would generate p_1_2.

I will add the separator anyway, as it's more unlikely than without separator.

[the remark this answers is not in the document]
cyenyxe: Yeah, I appreciate it is not a perfect solution :disappointed:

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/configuration/SubmittedVariantAccessioningConfiguration.java
```diff
@@ -64,8 +68,9 @@ public SubmittedVariantAccessioningDatabaseService submittedVariantAccessioningD
     }
 
     @Bean
-    public MonotonicAccessionGenerator<SubmittedVariantModel> submittedVariantAccessionGenerator() {
+    public MonotonicAccessionGenerator<ISubmittedVariant> submittedVariantAccessionGenerator() {
         ApplicationProperties properties = applicationProperties();
+        logger.debug("Using application properties: " + properties.toString());
```
cyenyxe: These should be reported by Spring Boot anyway, could you please check that they are not the same?

eva-accession-core/src/test/resources/accession-test.properties
```diff
@@ -0,0 +1,3 @@
+accessioning.instanceId=test-instance-01;
```
cyenyxe: Is the semicolon intentional?

eva-accession-ws/src/test/resources/accession-ws-test.properties
```diff
@@ -0,0 +1,3 @@
+accessioning.instanceId=test-ws-instance-01;
```
cyenyxe: Is the semicolon intentional?

eva-accession-core/src/test/resources/accession-test.properties
jmmut: hum, that came from accession-commons, I didn't realise. 

pika.. err.. @jorizci! I choose you!

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/configuration/SubmittedVariantAccessioningConfiguration.java
jmmut: couldn't find the values of the properties in the logs, even with boot at "debug" log level.

The only messages I found are similar to 
> Adding PropertySource 'applicationConfig: [classpath:/application.properties]' [...]
> Adding PropertySource 'class path resource [accession-test.properties]' [...]

which just says everything spring found, and 
> Retrieved ApplicationContext [...]  propertySourceLocations = '{classpath:accession-test.properties}' [...]

which just re-states what the annotation `@TestPropertiesSources` has in our code. 

Personally didn't find that information useful to make sure the properties taken were the expected ones. Actually, at first I used `@PropertiesSources` and spring didn't use the expected values.

jmmut:
please note that I added a small test in core, to make sure the configuration was correct, and I had to  remove the shared test configuration in core and the test-jar

eva-accession-core/src/test/resources/accession-test.properties
jorizci: Remove it, I was probably smoking a cookie or something similar

PR #8, opened 2018-04-06
EVA-1169 Write submitted variants to database

jmmut:


eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriter.java
```diff
@@ -15,10 +15,24 @@
...
-import uk.ac.ebi.eva.accession.core.persistence.SubmittedVariantEntity;
+import uk.ac.ebi.eva.accession.core.ISubmittedVariant;
+import uk.ac.ebi.eva.accession.core.SubmittedVariantAccessioningService;
 
-public class SummaryFileWriter extends CompositeItemWriter<SubmittedVariantEntity> {
-    // TODO: implement
+import java.util.List;
+import java.util.Map;
+
+public class AccessionWriter implements ItemWriter<ISubmittedVariant> {
```
cyenyxe: If this class consumes `ISubmittedVariant` objects, does that mean we can change the processor to have the interface as output type?

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriter.java
```diff
@@ -15,10 +15,24 @@
...
+
+    private SubmittedVariantAccessioningService service;
+
+    public AccessionWriter(SubmittedVariantAccessioningService service) {
+        this.service = service;
+    }
+
+    @Override
+    public void write(List<? extends ISubmittedVariant> variants) throws Exception {
+        Map<Long, ISubmittedVariant> accessions = service.getOrCreateAccessions(variants);
```
cyenyxe: For auditing purposes, it was agreed that the database entities would have a creation date. Can you please modify it the class accordingly?

Also, shall we change `SubmittedVariantEntity` it to be a `@Document` instead of an `@Entity`, or shall we wait until the benchmarks are finished? If we decide to wait we have to leave the JIRA ticket open.

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriterTest.java
```diff
@@ -0,0 +1,70 @@
...
+    public void saveSingleAccession() throws Exception {
+        SubmittedVariant variant = new SubmittedVariant("accession", TAXONOMY, "project", "contig", 100, "reference",
+                                                         "alternate", false);
+
+        accessionWriter.write(Collections.singletonList(variant));
+
+        Map<Long, ISubmittedVariant> accessions = service.getAccessions(Collections.singletonList(variant));
+        assertEquals(1, accessions.size());
+
+        ISubmittedVariant savedVariant = accessions.values().iterator().next();
```
cyenyxe: The value of the accession should be also checked.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriter.java
jmmut: yes, we can. it doesn't matter.
`public SimpleStepBuilder<I, O> processor(org.springframework.batch.item.ItemProcessor<? super I, ? extends O> processor)`

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriterTest.java
jmmut: added a check with the expectedId being `0`. Should we care here about the initial gap of 1 billion?

jmmut: agreed to put the first value as 10 billion

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriter.java
jmmut: I'm adding the the date for testing purposes until it's done in accession-commons, but we won't add the mongo part in this PR as agreed.

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriterTest.java
```diff
@@ -61,9 +63,10 @@ public void setUp() throws Exception {
     }
 
     @Test
+    @DirtiesContext
     public void saveSingleAccession() throws Exception {
         SubmittedVariant variant = new SubmittedVariant("accession", TAXONOMY, "project", "contig", 100, "reference",
```
cyenyxe: If the taxonomy is extracted into a constant to make the parameters more readable, I think the position should also be. In fact they are implemented that way in `VariantProcessorTest`. Applies to all the tests in this class.

jmmut: that is already done in another branch

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariant.java
```diff
@@ -149,12 +163,13 @@ public boolean equals(Object o) {
                 Objects.equals(projectAccession, that.projectAccession) &&
                 Objects.equals(contig, that.contig) &&
                 Objects.equals(referenceAllele, that.referenceAllele) &&
-                Objects.equals(alternateAllele, that.alternateAllele);
+                Objects.equals(alternateAllele, that.alternateAllele) &&
+                Objects.equals(createdDate, that.createdDate);
```
pabarcgar: Are we going to do any comparison of SubmittedVariant objects anywhere in the code? I was thinking in this possible problem:
- We process a VCF and we store the accession in the DB. 
- We process the same VCF again (maybe it wasn't complete the first time, maybe is a different process that takes a VCF and return the accessions from the DB...)
- We compare both objects and are different because the one read from the VCF has created date null. But they are really the same submitted variant.

If something like that could happen, then we should remove created date from equals and toHash.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariant.java
```diff
@@ -37,8 +38,11 @@
 
     private boolean supportedByEvidence;
 
+    private LocalDateTime createdDate;
+
     public SubmittedVariant(String assemblyAccession, int taxonomyAccession, String projectAccession, String contig,
-                            long start, String referenceAllele, String alternateAllele, boolean supportedByEvidence) {
+                            long start, String referenceAllele, String alternateAllele, boolean supportedByEvidence,
+                            LocalDateTime createdDate) {
```
pabarcgar: It looks that so far createdDate is always initialized as null. If there is still code to be added and that parameter is going to be used, I would leave it in the constructor. If not, I would remove it from the constructor, even if we keep getCreatedDate in the class just to implement the interface.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariant.java
jmmut: They won't be compared like this to decide about its insertion (or not insertion). For that there's this class, that selects the fields to identify the object https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariantModelSummaryFunction.java

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariant.java
jmmut: yeah, I was not convinced about this either. the class needs to have this field to be compliant with the interface, but maybe the setter is enough.

PR #12, opened 2018-04-18
EVA-1177 Configuration for API deployment

cyenyxe:
Spring Boot configuration and properties files for REST API running in Tomcat.

`accessioning-commons` dependency migrated from snapshot to release version.

### DIFF
--- a/eva-accession-pipeline/src/main/resources/application.properties
+++ b/eva-accession-pipeline/src/main/resources/application.properties
@@ -1,15 +1,10 @@
 spring.batch.job.names=
 
-spring.datasource.driver-class-name=org.postgresql.Driver
-spring.datasource.url=
-spring.datasource.username=
-spring.datasource.password=
-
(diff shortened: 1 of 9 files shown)


eva-accession-pipeline/src/main/resources/application.properties
```diff
@@ -20,4 +15,13 @@ parameters.vcfAggregation=
...
+parameters.forceRestart=
+
+
+spring.datasource.driver-class-name=org.postgresql.Driver
+spring.datasource.url=
+spring.datasource.username=
+spring.datasource.password=
+
+# Only to set up the database!
+# spring.jpa.generate-ddl=true
```
cyenyxe: To be honest I prefer the granularity `spring.jpa.hibernate.ddl-auto` provides, because we can set it to "validate" so it checks the schema every time the application is started.

PR #13, opened 2018-04-20
EVA-1181 Store accessioned variants in MongoDB instead of an RDBMS

jmmut:


eva-accession-core/pom.xml
```diff
@@ -25,6 +29,13 @@
             <groupId>uk.ac.ebi.ampt2d</groupId>
             <artifactId>accessioning-commons</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>com.github.fakemongo</groupId>
+            <artifactId>fongo</artifactId>
+            <scope>test</scope>
+            <version>${com.github.fakemongo.version}</version>
```
cyenyxe: You don't need to do this: specify the dependency in the parent like [this](https://github.com/EBIvariation/eva-accession/blob/develop/pom.xml#L29-L37) and it will only be inherited by those modules that specify it. I verified it yesterday after we discussed it at the office, using `mvn dependency:tree`.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/SubmittedVariantEntity.java
```diff
@@ -142,4 +131,25 @@ public boolean isSupportedByEvidence() {
...
+
+    @Override
+    public Long getId() {
+        return accession;
+    }
+
+    /**
+     * If we want to insert entities with the @Id already filled, this is needed for using @CreatedDate.
+     *
+     * Even If we implement the "save" method in the repository to avoid updates and make inserts always, in
```
cyenyxe: Even **if** we implement the "**insert**" method

eva-accession-core/src/test/resources/accession-test.properties
```diff
@@ -1,3 +1,5 @@
 accessioning.instanceId=test-instance-01
```
cyenyxe: Should we rename this file as `application.properties` until we have the need for multiple test properties?

jmmut: I did this to avoid confusions with the test `.properties` of other submodules

PR #18, opened 2018-05-17
EVA-1204 Post-accessioning validation checks (1/2)

jmmut:
first part of the ticket, don't move to done after merging this.

### DIFF
--- a/eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriter.java
+++ b/eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriter.java
@@ -47,34 +47,34 @@ public class AccessionWriter implements ItemStreamWriter<ISubmittedVariant> {
     public void write(List<? extends ISubmittedVariant> variants) throws Exception {
         List<AccessionWrapper<ISubmittedVariant, String, Long>> accessions = service.getOrCreateAccessions(variants);
         accessionReportWriter.write(accessions);
-        assertCountsMatch(variants, accessions);
+        checkCountsMatch(variants, accessions);
     }
 
-    void assertCountsMatch(List<? extends ISubmittedVariant> variants,
-                                   List<AccessionWrapper<ISubmittedVariant, String, Long>> accessions) {
+    void checkCountsMatch(List<? extends ISubmittedVariant> variants,
+                          List<AccessionWrapper<ISubmittedVariant, String, Long>> accessions) {
         if (variants.size() != accessions.size()) {
             Set<ISubmittedVariant> accessionedVariants = accessions.stream()
                                                                    .map(AccessionWrapper::getData)
                                                                    .collect(Collectors.toSet());
             HashSet<ISubmittedVariant> distinctVariants = new HashSet<>(variants);
             int duplicateCount = variants.size() - distinctVariants.size();
-
-            List<ISubmittedVariant> variantsWithoutAccession = distinctVariants.stream()
-                                                                               .filter(v -> !accessionedVariants.contains(v))
-                                                                               .collect(Collectors.toList());
             if (duplicateCount != 0) {
                 logger.warn("A variant chunk contains {} repeated variants. This is not an error, but please check " +
                                     "it's expected.", duplicateCount);
             }
+
+            List<ISubmittedVariant> variantsWithoutAccession = distinctVariants.stream()
+                                                                               .filter(v -> !accessionedVariants.contains(v))
+                                                                               .collect(Collectors.toList());
             if (variantsWithoutAccession.size() != 0) {
-                logger.error("A variant chunk wasn't accessioned properly. Only {} variants were accessioned out of " +
-                                     "{} distinct variants (from a chunk of {} variants, having {} repeated variants). " +
-                                     "The {} variants that were not accessioned are these: {}",
-                             accessionedVariants.size(), distinctVariants.size(), variants.size(), duplicateCount,
-                             variantsWithoutAccession.size(), variantsWithoutAccession.toString());
-                logger.info("This is the complete chunk of {} variants: {}", variants.size(), variants.toString());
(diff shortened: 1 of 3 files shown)


eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriterTest.java
```diff
@@ -188,4 +193,28 @@ public void createAccessionAndItAppearsInTheReportVcf() throws Exception {
...
+    public void shouldThrowIfSomeVariantsWereNotAccessioned() {
+        SubmittedVariant variant = new SubmittedVariant("assembly", TAXONOMY, "project", "contig", START_1,
+                                                        REFERENCE_ALLELE, ALTERNATE_ALLELE, false);
+
+        thrown.expect(IllegalStateException.class);
+        accessionWriter.assertCountsMatch(Collections.singletonList(variant), new ArrayList<>());
+    }
+
+    @Test
+    public void shouldThrowIfSomeVariantsWereNotAccessionedInAChunkWithRepeatedVariants() {
```
pabarcgar: can you add a test similar to this, with repeated variants, but where all of them have been accessioned and no exception is thrown? If that case is already covered in previous tests, then is not needed.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriter.java
```diff
@@ -39,6 +47,36 @@ public AccessionWriter(SubmittedVariantAccessioningService service, AccessionRep
...
+            Set<ISubmittedVariant> accessionedVariants = accessions.stream()
+                                                                   .map(AccessionWrapper::getData)
+                                                                   .collect(Collectors.toSet());
+            HashSet<ISubmittedVariant> distinctVariants = new HashSet<>(variants);
+            int duplicateCount = variants.size() - distinctVariants.size();
+
+            List<ISubmittedVariant> variantsWithoutAccession = distinctVariants.stream()
+                                                                               .filter(v -> !accessionedVariants.contains(v))
+                                                                               .collect(Collectors.toList());
+            if (duplicateCount != 0) {
```
pabarcgar: I think this check and message should be before the previous statement.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriter.java
```diff
@@ -39,6 +47,36 @@ public AccessionWriter(SubmittedVariantAccessioningService service, AccessionRep
...
+                logger.warn("A variant chunk contains {} repeated variants. This is not an error, but please check " +
+                                    "it's expected.", duplicateCount);
+            }
+            if (variantsWithoutAccession.size() != 0) {
+                logger.error("A variant chunk wasn't accessioned properly. Only {} variants were accessioned out of " +
```
pabarcgar: this is a complicated message, but I guess that is not going to appear too much, and when it does we want as much detail as possible, so I think this is actually OK.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriter.java
```diff
@@ -39,6 +47,36 @@ public AccessionWriter(SubmittedVariantAccessioningService service, AccessionRep
...
+                                     "{} distinct variants (from a chunk of {} variants, having {} repeated variants). " +
+                                     "The {} variants that were not accessioned are these: {}",
+                             accessionedVariants.size(), distinctVariants.size(), variants.size(), duplicateCount,
+                             variantsWithoutAccession.size(), variantsWithoutAccession.toString());
+                logger.info("This is the complete chunk of {} variants: {}", variants.size(), variants.toString());
```
pabarcgar: How big is a batch by default? This could be an impractical message.

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriterTest.java
jmmut: `saveSameAccessionTwice` checks a repeated variant is accessioned correctly, calling `AccessionWriter::write`, and this calls `AccessionWriter::assertCountsMatch`. So it's tested at class level, not method level. should I add another test?

pabarcgar: As is already tested I don't think is needed then.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriter.java
cyenyxe: A suggestion to to summarize the report a bit more while (I think) keeping it understandable:

> A problem occurred while accessioning a chunk. Size = {variants.size}, distinct = {distinctVariants.size}, duplicate = {duplicateCount}, accessioned = {accessionedVariants.size}, not accessioned = {variantsWithoutAccession.size}
> 
> The non-accessioned variants are: {variantsWithoutAccession.toString}

It allows to grep easily only the counts, or both the counts + list of variants, which is in the next line to make the formatting a bit nicer.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriter.java
```diff
@@ -39,6 +47,36 @@ public AccessionWriter(SubmittedVariantAccessioningService service, AccessionRep
...
+                throw new IllegalStateException(
+                        "A variant chunk wasn't accessioned properly (the relevant information was already logged).");
```
cyenyxe: `A problem occurred while accessioning a chunk. See log for details.`

PR #19, opened 2018-05-18
EVA-1204 Post-accessioning validation checks (2/2)

jmmut:
requires https://github.com/EBIvariation/variation-commons/pull/71

I still have to do a manual run to check that everything is wired properly, but the review can already be done.

Note that there are several ways the VCF comparison could be done. The one here tries to avoid loading the whole input VCF in memory, by loading only a chunk. In most executions this will be enough, but if the report VCF is unordered in the absolute worst way, the report VCF may be completely loaded in memory.


jmmut:
forgot that some changes were needed in variation-commons: https://github.com/EBIvariation/variation-commons/pull/71

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/CoordinatesVcfFactory.java
```diff
@@ -0,0 +1,80 @@
...
+ * Transforms a String (in VCF format) to a list of variants, keeping only the basic fields (coordinates):
+ * chromosome, position, id, reference, alternate.
+ */
+public class CoordinatesVcfFactory extends VariantVcfFactory {
+
+    @Override
+    public List<Variant> create(String fileId, String studyId, String line)
+            throws IllegalArgumentException, NonVariantException, IncompleteInformationException {
+        String[] fields = line.split("\t", 6);
+        String chromosome = getChromosomeWithoutPrefix(fields);
```
andresfsilva: The methods `getChromosomeWithoutPrefix`, `getPosition`, `getReference`, `getAlternateAlleles` from `VariantVcfFactory` class can't be used here because they are private and that is why we are getting the compilation error.

Methods could be changed to protected in the commons project if they need to be used in the subclass.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/configuration/jobs/steps/CheckSubsnpAccessionsStepConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+
+@Configuration
+@EnableBatchProcessing
+public class CheckSubsnpAccessionsStepConfiguration {
+
+    @Autowired
+    private InputParameters inputParameters;
+
+    @Autowired
+    private VcfReader vcfReader;
```
pabarcgar: Is this VcfReader autowired with the input VCF?

jmmut: I just added a qualifier

pabarcgar: 👍

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/ReportCheckTasklet.java
```diff
@@ -0,0 +1,235 @@
...
+            }
+        }
+        accessionBuffer.removeAll(matchedAccessions);
+    }
+
+    private boolean readAccessionsAndRemoveMatchingOnes(Set<Variant> variantBuffer, Set<Variant> accessionBuffer)
+            throws Exception {
+        Set<Variant> newUnmatchedAccessions = new HashSet<>();
+        List<Variant> reportVariantsRead = null;
+        newUnmatchedAccessions.clear();
```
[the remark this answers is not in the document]
pabarcgar: Were you trying to avoid creating a new set every time this method was called? Apparently, the performance of those two options is not so different and they recommend to use the one that allows a code easier to refactor. In this case, I would say that creating the set instead of reusing and clear one is cleaner.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/ReportCheckTasklet.java
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * To perform self-checks, this tasklet provides the maximum size that the accessionBuffer during the execution, and
+ * also provides the number of iterations needed.
+ *
+ */
+public class ReportCheckTasklet implements Tasklet {
+
+    private static final Logger logger = LoggerFactory.getLogger(ReportCheckTasklet.class);
+
+    private static final int DEFAULT_MAX_BUFFER_SIZE = 100000;
```
cyenyxe: Use `chunk_size * 2`? It would make sense to keep it relatively close to the number of records read at once from the input VCF file.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/ReportCheckTasklet.java
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * Terminology: The original VCF contains variants, and it will be loaded into the 'variantBuffer'. The accession
+ * report VCF contains the accessioned variants, and it will be loaded into the 'accessionBuffer'.
+ *
+ * The high-level idea of the process is filling those 2 buffers, and when the same coordinates appears in both the
+ * variantBuffer and the accessionBuffer, it means that the variant in those coordinates was correctly accessioned,
+ * so it can be removed from both buffers. At the end, only variants without accessions and accessions without
+ * variants should be left in the buffers after both files were completely read.
+ *
+ * The reason why the buffers are needed is that the accession report can be unordered. In order to avoid
```
cyenyxe: Given that we are writing small chunks (not like in the RS release process), we could sort the chunk right before appending to the report. That way we don't need to call an external sorting tool after the accessioning process is completed, it should simplify the testing a bit, and we can keep the buffer size smaller.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/ReportCheckTasklet.java
```diff
@@ -0,0 +1,235 @@
...
+    /**
+     * See the terminology in the documentation of this class
+     * @param maxVariantBufferSize max size of the buffer where the variants of the original VCF are loaded
+     */
+    public void setMaxVariantBufferSize(long maxVariantBufferSize) {
+        this.maxVariantBufferSize = maxVariantBufferSize;
+    }
+
+    /**
+     * To grant correct behaviour in worst-ordering scenario, the accessionBuffer (for yet-unmatched accessions from
```
cyenyxe: * "guarantee" instead of "grant"
* "accession buffer" instead of "accessionBuffer" to allow the documentation to still stand if the variable name changes

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/ReportCheckTasklet.java
```diff
@@ -0,0 +1,235 @@
...
+        this.iterations = 0;
+        this.skipPolicy = new InvalidVariantSkipPolicy();
+    }
+
+    @Override
+    public RepeatStatus execute(StepContribution stepContribution, ChunkContext chunkContext) throws Exception {
+        Set<Variant> variantBuffer = new HashSet<>();
+        Set<Variant> accessionBuffer = new HashSet<>();
+        vcfReader.open(new ExecutionContext());
+        reportReader.open(new ExecutionContext());
```
cyenyxe: Maybe `inputReader` and `inputBuffer` instead of `vcfReader` (because the report is also a VCF) and `variantBuffer` (because the report also contains variants).

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/ReportCheckTaskletTest.java
```diff
@@ -0,0 +1,171 @@
...
+                new StepExecution(CHECK_SUBSNP_ACCESSION_STEP, new JobExecution(JOB_ID)));
+        reportCheckTasklet.execute(stepContribution, null);
+
+        // then
+        assertEquals(ExitStatus.COMPLETED, stepContribution.getExitStatus());
+    }
+
+    private ReportCheckTasklet getReportCheckTasklet(URI vcfUri, URI reportUri) throws IOException {
+        File vcfFile = new File(vcfUri);
+        AggregatedVcfReader vcfReader = new AggregatedVcfReader("fileId", "studyId", Aggregation.BASIC, null, vcfFile);
```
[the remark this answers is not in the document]
jmmut: yes, the test `originalVcfContainsNonVariants` is done on a genotyped vcf

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/ReportCheckTasklet.java
```diff
@@ -0,0 +1,235 @@
...
+        }
+        return variants;
+    }
+
+    private void removeMatchingVariants(Set<Variant> variantBuffer, Set<Variant> accessionBuffer) {
+        Set<Variant> matchedAccessions = new HashSet<>();
+        for (Variant unmatchedAccession : accessionBuffer) {
+            boolean removed = variantBuffer.remove(unmatchedAccession);
+            if (removed) {
+                matchedAccessions.add(unmatchedAccession);
```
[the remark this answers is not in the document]
jmmut: it's usually a bad idea to modify a collection while you are traversing it. the correct way to do that would be with iterators:
```
    Iterator<Variant> accessionBufferIterator = accessionBuffer.iterator();
    while(accessionBufferIterator.hasNext()) {
        Integer unmatchedAccession = accessionBufferIterator.next();
        boolean removed = variantBuffer.remove(unmatchedAccession);
        if (removed) {
            accessionBufferIterator.remove();
        }
        if (variantBuffer.isEmpty()) {
            break;
        }
    }
```
which I don't mind to do, but someone might try to "modernize" the Iterator+while to a forEach (`for( ... : ...)`) and that would be wrong. see https://stackoverflow.com/a/7393539/2375586

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/ReportCheckTasklet.java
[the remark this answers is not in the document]
jmmut: this version resulted after lots of refactorings. indeed, now it's useless to do a `.clear` just after a new instance is created. I'll remove the `clear`.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/CoordinatesVcfFactory.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Transforms a String (in VCF format) to a list of variants, keeping only the basic fields (coordinates):
+ * chromosome, position, id, reference, alternate.
+ */
+public class CoordinatesVcfFactory extends VariantVcfFactory {
```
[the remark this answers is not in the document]
jmmut: I don't think so because it wouldn't reject non-variants and would accession them

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/ReportCheckTasklet.java
jmmut: making the buffer size a function of the chunk size is a good idea. if the inter-chunk order is maintained (which I guess is what spring batch wihout concurrency does), chunk_size would be enough. But the performance of the algorithm quickly degrades if the buffer is not enough due to unorderings. Even though in theory we shouldn't get those inter-chunk unorderings, `chunk_size * 2` feels a bit tight, but as I can't provide real arguments, we can use that.

jmmut: note that this is the default, and we'll use a setter when we inject the tasklet bean

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/ReportCheckTasklet.java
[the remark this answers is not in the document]
pabarcgar: You are right! 👍  I prefer the "for" here

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriterTest.java
```diff
@@ -174,4 +183,43 @@ private void assertHeaderIsNotWrittenTwice(File output) throws IOException {
...
+        accessionReportWriter.write(mockWrap(variants));
+
+        // then
+        BufferedReader fileInputStream = new BufferedReader(new InputStreamReader(new FileInputStream(output)));
+        String line;
+        while ((line = fileInputStream.readLine()) != null) {
+            if (!line.startsWith("#")) {
+                break;
+            }
+        }
```
cyenyxe: You could do just:

```java
while ((line = fileInputStream.readLine()) != null && line.startsWith("#"));
```

This is fine because Java [guarantees sub-expressions to be evaluated from left to right](https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.7).

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/AccessionWrapperComparatorTest.java
```diff
@@ -0,0 +1,90 @@
...
+import uk.ac.ebi.eva.accession.core.SubmittedVariant;
+
+import static org.junit.Assert.assertEquals;
+
+public class AccessionWrapperComparatorTest {
+
+    @Test
+    public void checkSortedContig() {
+        AccessionWrapperComparator accessionWrapperComparator = new AccessionWrapperComparator();
+        int comparation = accessionWrapperComparator.compare(buildMockAccessionWrapper("A", 200),
```
cyenyxe: `comparison` instead of `comparation`

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/configuration/jobs/steps/CheckSubsnpAccessionsStepConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+    @Qualifier(REPORT_READER)
+    public ItemStreamReader<Variant> reportReader() throws IOException {
+        VcfReader vcfReader = new VcfReader(new CoordinatesVcfLineMapper(), new File(inputParameters.getOutputVcf()));
+        return new UnwindingItemStreamReader<>(vcfReader);
+    }
+
+    @Bean(CHECK_SUBSNP_ACCESSION_STEP)
+    public Step checkSubsnpAccessionStep(StepBuilderFactory stepBuilderFactory) throws IOException {
+        ReportCheckTasklet tasklet = new ReportCheckTasklet(inputReader, reportReader());
+        tasklet.setInitialBufferSize(inputParameters.getChunkSize() * 2);
```
cyenyxe: Why not providing it in the constructor? It is not like we have other uses for this so we don't need to over-engineer it.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/ReportCheckTasklet.java
```diff
@@ -0,0 +1,201 @@
...
+
+    @Override
+    public RepeatStatus execute(StepContribution stepContribution, ChunkContext chunkContext) throws Exception {
+        Set<Variant> variantBuffer = new HashSet<>();
+        Set<Variant> accessionBuffer = new HashSet<>();
+        inputReader.open(new ExecutionContext());
+        reportReader.open(new ExecutionContext());
+        boolean finishedInputFile = false;
+        boolean finishedReportFile = false;
+        while (!(finishedInputFile && finishedReportFile)) {
```
cyenyxe: This condition could be simplified if the meaning of the variables was inverted, as in `pendingInputFile`:

```java
boolean pendingInputFile = true;
boolean pendingReportFile = true;
while (pendingInputFile || pendingReportFile) {
    // ...
}
```

The return value in `fillbuffer` would then be `true` if more data can be read.

I don't have a strong opinion on this though.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriter.java
```diff
@@ -23,6 +23,7 @@
 import uk.ac.ebi.ampt2d.commons.accession.core.AccessionWrapper;
 import uk.ac.ebi.eva.accession.core.ISubmittedVariant;
 import uk.ac.ebi.eva.accession.core.SubmittedVariant;
+import uk.ac.ebi.eva.accession.pipeline.steps.tasklets.reportCheck.AccessionWrapperComparator;
```
pabarcgar: Import not used

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/ReportCheckTasklet.java
pabarcgar: I like a little bit more "while (pendingInputFile || pendingReportFile)"
In the other hand, if I see that fillBuffer() returns true, I tend to think more that it has finished. But the variable name in the side will clarify that. Looking one or the other you know what fillBuffer is returning:
finishedInputFile = fillBuffer(inputReader, variantBuffer);
vs 
pendingInputFile = fillBuffer(inputReader, variantBuffer);

pabarcgar: I would change it then, but is OK also like it is

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/AccessionWrapperComparator.java
```diff
@@ -0,0 +1,74 @@
...
+
+public class AccessionWrapperComparator implements Comparator<AccessionWrapper<ISubmittedVariant, String, Long>> {
+
+    private Map<String, Integer> contigOrder;
+
+    public AccessionWrapperComparator(List<? extends ISubmittedVariant> variants) {
+        this.contigOrder = getContigOrder(variants);
+    }
+
+    private Map<String, Integer> getContigOrder(List<? extends ISubmittedVariant> variants) {
```
pabarcgar: I think that iterating all the variants just to get the contig order is a little bit too much. In the other hand, they will be the  variants in a single Batch that is already in memory, so it shouldn't take long and the impact of this will be minimal. And this is being made to improve the performance of the checking algorithm, so probably the penalty is paid off later.

If we see this validation is too slow when we start testing with real data, we can take a look at this, otherwise, is OK like it is.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/ReportCheckTasklet.java
```diff
@@ -150,46 +149,23 @@ private void removeMatchingVariants(Set<Variant> variantBuffer, Set<Variant> acc
...
-                if (!removed) {
-                    variantBuffer.add(variantWithoutAccession);
-                }
-            }
-        }
+    /**
+     * TODO: decide if we allow to grow indefinitely or put a hard cap
+     * A hard cap would mean we stop the process if there are more unmatched variants than the buffer size or the
+     * report is heavily unordered.
+     * Alternatively, we can allow configuring which behaviour to use?
```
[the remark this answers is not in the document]
jmmut: discussed offline, the mismatching variants are reported only at the end, but with the new strategy of sorting according to the input VCF, it should only grow if something with the accessioning went really bad and there are lots of mismatching variants

PR #23, opened 2018-06-18
EVA-1208 EVA-1235 Fix submitted variant model summary function

cyenyxe:
The EVA groups variants based on the reference assembly they are mapped against. A species and its subspecies are considered under a single group, so the taxonomy needs to be removed from the summary function. dbSNP SubSNPs do take into account both.

### DIFF
--- a/eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantModelSummaryFunctionTest.java
+++ b/eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantModelSummaryFunctionTest.java
@@ -1,0 +1,152 @@
+/*
+ *
+ * Copyright 2018 EMBL - European Bioinformatics Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantModelSummaryFunctionTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+    private SubmittedVariantModelSummaryFunction summaryFunction;
+
+    @Before
+    public void setUp() {
+        summaryFunction = new SubmittedVariantModelSummaryFunction();
+    }
+
+    @Test
+    public void summaryFunctionMustBeIdempotent() {
```
jmmut: I don't think idempotent is the right word here. idempotence means `f(x) == f(f(x))`

cyenyxe: https://en.wikipedia.org/wiki/Idempotence#Computer_science_meaning

jmmut: alright then. not my favorite thing to do, though, taking concise words and use them with a different meaning. Damn computer scientist! they ruined computer science!

PR #26, opened 2018-06-28
EVA-1262 Overlapping SNPs and INDELs are not properly sorted in output report

andresfsilva:
Denormilize accessions before sorting instead of doing it after


eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriter.java
```diff
@@ -119,14 +119,13 @@ public void write(List<? extends AccessionWrapper<ISubmittedVariant, String, Lon
...
-        ISubmittedVariant variant = denormalizeVariant(normalizedVariant);
-        String vcfLine = variantToVcfLine(id, variant);
+    private void writeVariant(Long id, ISubmittedVariant denormalizedVariant) throws IOException {
+        String vcfLine = variantToVcfLine(id, denormalizedVariant);
         fileWriter.write(vcfLine);
         fileWriter.newLine();
     }
 
-    private ISubmittedVariant denormalizeVariant(ISubmittedVariant normalizedVariant) {
+    protected ISubmittedVariant denormalizeVariant(ISubmittedVariant normalizedVariant) {
```
cyenyxe: This method is no longer used in this class but in `AccessionWriter`, so I think it is more appropriate to move there this and other methods it depends on, and keep it private.

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/configuration/jobs/CreateSubsnpAccessionsJobConfigurationTest.java
```diff
@@ -66,6 +69,30 @@ public void executeJob() throws Exception {
...
+        JobExecution jobExecution = jobLauncherTestUtils.launchJob();
+        assertEquals(BatchStatus.COMPLETED, jobExecution.getStatus());
+
+        File output = new File(inputParameters.getOutputVcf());
+        int firstVariantLineNumber = getVariantLineByPosition(output, "20" + "\t" + "60027");
+        int secondVariantLineNumber = getVariantLineByPosition(output, "20" + "\t" + "60028");
+        assertTrue(firstVariantLineNumber < secondVariantLineNumber);
+    }
+
+    public static int getVariantLineByPosition(File output, String position) throws IOException {
```
cyenyxe: `getVariantLineNumberByPosition` for complete clarity

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriter.java
```diff
@@ -47,8 +48,14 @@ public AccessionWriter(SubmittedVariantAccessioningService service, AccessionRep
...
-        accessions.sort(new AccessionWrapperComparator(variants));
-        accessionReportWriter.write(accessions);
+
+        List<AccessionWrapper<ISubmittedVariant, String, Long>> denormalizedAccessions = new ArrayList<>();
+        accessions.forEach(accession -> denormalizedAccessions
+                .add(new AccessionWrapper(accession.getAccession(), accession.getHash(),
+                                          accessionReportWriter.denormalizeVariant(accession.getData()))));
+
+        denormalizedAccessions.sort(new AccessionWrapperComparator(variants));
+        accessionReportWriter.write(denormalizedAccessions);
```
cyenyxe: I know it was already this way, but given that the sorting is specific to the report, maybe it should be called first thing inside the `write` method?

eva-accession-pipeline/src/test/resources/accession-pipeline-test.properties
```diff
@@ -8,7 +8,7 @@ parameters.assemblyAccession=assembly
 parameters.taxonomyAccession=1111
 parameters.projectAccession=project
 parameters.chunkSize=100
-parameters.vcf=src/test/resources/input-files/vcf/genotyped.vcf.gz
+parameters.vcf=src/test/resources/input-files/vcf/small_genotyped.vcf.gz
```
cyenyxe: A header line indicating that one record is a non-variant would be useful, because it took me a while to understand why only 22 out of 23 were expected to be accessioned.

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/configuration/jobs/CreateSubsnpAccessionsJobConfigurationTest.java
```diff
@@ -66,6 +69,30 @@ public void executeJob() throws Exception {
...
+        BufferedReader fileInputStream = new BufferedReader(new InputStreamReader(new FileInputStream(output)));
+        String line;
+        int lineNumber = 0;
+        while ((line = fileInputStream.readLine()) != null) {
+            if (line.startsWith(position)) {
+                return lineNumber;
+            }
+            lineNumber++;
+        }
+        throw new IllegalStateException("VCF didn't have any data lines");
```
pabarcgar: This message is not accurate. If the VCF has data lines, but the "position" we are looking is not in the VCF, the exception will be thrown as well. Maybe this message is better:
"The VCF does not contain any variant with position " + position

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriterTest.java
```diff
@@ -156,6 +156,39 @@ public void saveTwoAccession() throws Exception {
...
+        BufferedReader fileInputStream = new BufferedReader(new InputStreamReader(new FileInputStream(output)));
+        String line;
+        int lineNumber = 0;
+        while ((line = fileInputStream.readLine()) != null) {
+            if (line.startsWith(position)) {
+                return lineNumber;
+            }
+            lineNumber++;
+        }
+        throw new IllegalStateException("VCF didn't have any data lines");
```
pabarcgar: The same comment that in this method in the other test:

This message is not accurate. If the VCF has data lines, but the "position" we are looking is not in the VCF, the exception will be thrown as well. Maybe this message is better:
"The VCF does not contain any variant with position " + position

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriter.java
andresfsilva: Denormalization and sorting was moved to this class (AccessionReportWriter) istead of having it in the AccessionWriter

PR #28, opened 2018-07-03
EVA-1218 Store SubSNP information in accessioning service

jmmut:


eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/SubmittedVariantEntity.java
```diff
@@ -47,22 +48,31 @@
...
+    @Indexed
+    @Field("rs")
+    private Long clusteredVariant;
+
     @Field("evidence")
     private boolean supportedByEvidence;
 
-    SubmittedVariantEntity() {
+    @Field("matchAsm")
+    private Boolean matchesAssembly;
```
jmmut: alternatively, we can move this field to ImportedSubmittedVariantEntity only and fill the getter in this class as:
```
    @Override
    public Boolean getMatchesAssembly() {
        return null;
    }
```
and continue assuming in other places that null means true

eva-accession-import/pom.xml
```diff
@@ -22,6 +22,10 @@
             <artifactId>eva-accession-core</artifactId>
             <version>0.1</version>
         </dependency>
+        <dependency>
+            <groupId>com.github.fakemongo</groupId>
+            <artifactId>fongo</artifactId>
+        </dependency>
```
cyenyxe: Given that Fongo already causes issues during testing and we are using a real Mongo database in other tests, do we really want to keep it for some particular classes?

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/ImportedSubmittedVariantWriterTest.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Function;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(SpringRunner.class)
+@ContextConfiguration(classes = {MongoConfiguration.class, MongoTestConfiguration.class})
+public class ImportedSubmittedVariantWriterTest {
```
cyenyxe: This differs from the way tests were written [for the accessioning pipeline](https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriterTest.java).

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/ImportedSubmittedVariantWriterTest.java
```diff
@@ -0,0 +1,159 @@
...
+                                                                      START_1, "reference", "alternate",
+                                                                      CLUSTERED_VARIANT, false, MATCHES_ASSEMBLY);
+        ImportedSubmittedVariantEntity variant = new ImportedSubmittedVariantEntity(
+                EXPECTED_ACCESSION, hashingFunction.apply(submittedVariant), submittedVariant);
+
+        thrown.expect(RuntimeException.class);
+        importedSubmittedVariantWriter.write(Arrays.asList(variant, variant));
+    }
+
+    private void assertVariantEquals(ISubmittedVariant expectedvariant, ISubmittedVariant actualVariant) {
```
cyenyxe: Implement `equals` and `hashCode` in the corresponding classes following [this model](https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariant.java), then include the extra fields to this method if you want to. But don't compare every single field and leave the classes with no `equals` method!

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionWriterTest.java
```diff
@@ -127,16 +131,20 @@ private void assertVariantEquals(ISubmittedVariant expectedvariant, ISubmittedVa
         assertEquals(expectedvariant.getStart(), actualVariant.getStart());
         assertEquals(expectedvariant.getReferenceAllele(), actualVariant.getReferenceAllele());
         assertEquals(expectedvariant.getAlternateAllele(), actualVariant.getAlternateAllele());
+        assertEquals(expectedvariant.getClusteredVariant(), actualVariant.getClusteredVariant());
         assertEquals(expectedvariant.isSupportedByEvidence(), actualVariant.isSupportedByEvidence());
+        assertEquals(expectedvariant.getMatchesAssembly(), actualVariant.getMatchesAssembly());
     }
```
cyenyxe: Same comment regarding the `equals` method as in `ImportedSubmittedVariantWriterTest.java`.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/SubmittedVariantEntity.java
jmmut: you mean "store it only when it **doesn't** match the assembly" right?

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/persistence/ImportedSubmittedVariantEntity.java
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.ac.ebi.eva.accession.dbsnp.persistence;
+
+import uk.ac.ebi.eva.accession.core.ISubmittedVariant;
+import uk.ac.ebi.eva.accession.core.persistence.SubmittedVariantEntity;
+
+public class ImportedSubmittedVariantEntity extends SubmittedVariantEntity {
```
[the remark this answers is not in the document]
jmmut: +1 to DbsnpSubmittedVariantEntity

eva-accession-import/pom.xml
jmmut: given that it's not that important for this module, let's not hurry and see if the fixed fongo gets deployed https://github.com/fakemongo/fongo/issues/343#issuecomment-402500418

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/SubmittedVariantEntity.java
jmmut: checked. it's not writtten, as we intend.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/ImportedSubmittedVariantWriterTest.java
jmmut: then I'm going to use the SummaryFunction to not replicate the selection of identifying fields

cyenyxe: Even if you use the summary function here, please still implement `equals` and `hashCode` methods in the corresponding classes. Those do not use the identifying fields like the summary function, but are pure equivalence functions.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/SubmittedVariantEntity.java
cyenyxe: > you mean "store it only when it doesn't match the assembly" right?

I was thinking only about the dbSNP variants, but given that all the EVA variants should match it, then you're right it's better to store it only when it doesn't match. Same for the alleles mismatch then.

The 'validated' flag is the only one that would be inconsistent, representing a positive instead of a negative trait, but again the number of validated ones is much lower so it makes sense in terms of storage optimization.

eva-accession-import/pom.xml
cyenyxe: I didn't know someone had recently addressed the issue, I'm fine with waiting then.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/SubmittedVariantEntity.java
pabarcgar: I agree with storing just the flags that are less usual (matchAssembly=false, validated=true ...)

andresfsilva: I agree, storing the less common ones would be a better option in order to optimise

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/summary/SubmittedVariantSummaryFunctionTest.java
```diff
@@ -39,20 +42,24 @@
 
     private static final String ALT_T = "T";
 
+    private static final Long CLUSTERED_VARIANT = null;
```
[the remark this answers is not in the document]
pabarcgar: well, thinking twice, I guess this is going to be done in separate tickets and PRs, so there is no need to do this now.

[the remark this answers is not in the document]
jmmut: No, I changed it in DbsnpSubmittedVariantSummaryFunctionTest but missed it in SubmittedVariantSummaryFunctionTest

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/SubmittedVariantEntity.java
jmmut: ok, so applying to every flag the rule of not writing for the most common value, this are the defaults (used if the flag is missing):
- supportedByEvidence true 
- matchesAssembly true
- allelesMatch true
- validated false

although the first one won't be usually true for dbsnp, but will be for eva. I wouldn't support different defaults depending of eva or dbsnp because it might require more engineering to avoid confusion.

cyenyxe: I agree with these defaults.

pabarcgar: 👍

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/summary/DbsnpSubmittedVariantSummaryFunctionTest.java
```diff
@@ -160,11 +179,12 @@ public void sameSummaryWhenClusteredVariantDiffers() {
...
         ISubmittedVariant submittedVariant1 =
                 new SubmittedVariant(ASSEMBLY_ACCESSION, TAXONOMY_ACCESSION, PROJECT_ACCESSION, CONTIG, START,
-                                     REF_A, ALT_T, CLUSTERED_VARIANT, SUPPORTED_BY_EVIDENCE, MATCHES_ASSEMBLY);
+                                     REF_A, ALT_T, CLUSTERED_VARIANT, SUPPORTED_BY_EVIDENCE, MATCHES_ASSEMBLY,
+                                     ALLELES_MATCH, VALIDATED);
 
         ISubmittedVariant submittedVariant2 =
                 new SubmittedVariant(ASSEMBLY_ACCESSION, TAXONOMY_ACCESSION, PROJECT_ACCESSION, CONTIG, START,
-                                     REF_A, ALT_T, CLUSTERED_VARIANT, false, MATCHES_ASSEMBLY);
+                                     REF_A, ALT_T, CLUSTERED_VARIANT, false, MATCHES_ASSEMBLY, ALLELES_MATCH, VALIDATED);
```
cyenyxe: Judging by the rest of declarations, it looks like this one goes over the 120 characters?

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/summary/SubmittedVariantSummaryFunctionTest.java
```diff
@@ -134,26 +149,88 @@ public void differentSummaryWhenReferenceDiffers() {
...
     public void sameSummaryWhenSupportedByEvidenceDiffers() {
         ISubmittedVariant submittedVariant1 =
                 new SubmittedVariant(ASSEMBLY_ACCESSION, TAXONOMY_ACCESSION, PROJECT_ACCESSION, CONTIG, START,
-                                     REF_A, ALT_T, CLUSTERED_VARIANT, SUPPORTED_BY_EVIDENCE, MATCHES_ASSEMBLY);
+                                     REF_A, ALT_T, CLUSTERED_VARIANT, SUPPORTED_BY_EVIDENCE, MATCHES_ASSEMBLY,
+                                     ALLELES_MATCH, VALIDATED);
+
+        ISubmittedVariant submittedVariant2 =
+                new SubmittedVariant(ASSEMBLY_ACCESSION, TAXONOMY_ACCESSION, PROJECT_ACCESSION, CONTIG, START,
+                                     REF_A, ALT_T, CLUSTERED_VARIANT, false, MATCHES_ASSEMBLY, ALLELES_MATCH, VALIDATED);
```
cyenyxe: Add line break to avoid going over 120 characters.

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriterTest.java
```diff
@@ -121,46 +128,39 @@ public static String getFirstVariantLine(File output) throws IOException {
...
     @Test
     public void writeInsertionWithAccession() throws IOException {
-        SubmittedVariant variant = new SubmittedVariant("accession", TAXONOMY, "project", CONTIG_1, START_1, "",
-                                                        ALTERNATE, CLUSTERED_VARIANT, false, MATCHES_ASSEMBLY);
+        writeIndeWithAccession("", ALTERNATE, CONTEXT_BASE, CONTEXT_BASE + ALTERNATE);
+    }
 
-        AccessionWrapper<ISubmittedVariant, String, Long> accessionWrapper =
-                new AccessionWrapper<ISubmittedVariant, String, Long>(ACCESSION, "1", variant);
+    private void writeIndeWithAccession(String reference, String alternate, String denormalizedReference,
```
cyenyxe: `writeIndelWithAccessionHelper` (to fix typo and differentiate more from the actual test methods)

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/SubmittedVariantEntity.java
```diff
@@ -81,9 +86,35 @@ public SubmittedVariantEntity(Long accession, String hashedMessage, String assem
...
-        this.supportedByEvidence = isSupportedByEvidence;
-        this.matchesAssembly = matchesAssembly;
+        this.clusteredVariantAccession = clusteredVariantAccession;
+        this.supportedByEvidence = supportedByEvidence == getDefaultSupportedByEvidence() ? null : supportedByEvidence;
+        this.matchesAssembly = matchesAssembly == getDefaultMatchesAssembly() ? null : matchesAssembly;
+        this.allelesMatch = allelesMatch == getDefaultAllelesMatch()? null : allelesMatch;
+        this.validated = validated == getDefaultValidated() ? null : validated;
+    }
+
+    public ISubmittedVariant toISubmittedVariant() {
```
cyenyxe: Implement new tests to check this method.

eva-accession-ws/src/test/java/uk/ac/ebi/eva/accession/ws/SubmittedVariantsRestControllerTest.java
```diff
@@ -58,28 +63,43 @@
...
     public void testGetVariantsRestApi() throws AccessionCouldNotBeGeneratedException {
-        List<AccessionResponseDTO> generatedAccessions = basicRestController.generateAccessions(
-                getListOfVariantMessages());
+        List<AccessionResponseDTO<SubmittedVariantDTO, ISubmittedVariant, String, Long>> generatedAccessions =
+                basicRestController.generateAccessions(getListOfVariantMessages());
         assertEquals(2, generatedAccessions.size());
         String accessions = generatedAccessions.stream().map(acc -> acc.getAccession().toString()).collect(
                 Collectors.joining(","));
         String getVariantsUrl = URL + accessions;
         ResponseEntity<List> getVariantsResponse = testRestTemplate.getForEntity(getVariantsUrl, List.class);
```
cyenyxe: If you declare this as `ResponseEntity<List<SubmittedVariantDTO>>` will ensure type safety and simplify the parsing, because you can access the objects instead of maps.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/ISubmittedVariant.java
```diff
@@ -41,14 +41,110 @@
...
+
+    /**
+     * @return True if the variant was curated manually, and not only detected by computational methods.
+     * see https://www.ncbi.nlm.nih.gov/books/NBK21088/table/ch5.ch5_t4/?report=objectonly
+     */
+    Boolean getValidated();
+
     LocalDateTime getCreatedDate();
+
+    static boolean equals(ISubmittedVariant first, ISubmittedVariant second) {
```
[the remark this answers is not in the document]
jmmut: agreed offline to 1) extract the method to another class so it's not confused as an overriding `default equals` method and 2) remove the DTO child and add a copy constructor in SubmittedVariant so that only there an autogenerated equals is needed.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/SubmittedVariantInactiveEntity.java
```diff
@@ -52,7 +53,7 @@
     private Boolean supportedByEvidence;
 
     @Field("matchAsm")
```
cyenyxe: `asmMatch`

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/utils/ISubmittedVariantComparator.java
```diff
@@ -0,0 +1,106 @@
...
+        if (!first.getAlternateAllele().equals(second.getAlternateAllele())) {
+            return false;
+        }
+        if (first.getClusteredVariantAccession() != null ? !first.getClusteredVariantAccession().equals(
+                second.getClusteredVariantAccession()) : second.getClusteredVariantAccession() != null) {
+            return false;
+        }
+        if (first.isSupportedByEvidence() != null ? !first.isSupportedByEvidence().equals(
+                second.isSupportedByEvidence()) : second.isSupportedByEvidence() != null) {
+
```
cyenyxe: Remove this blank line and another one below.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/SubmittedVariantInactiveEntity.java
```diff
@@ -18,30 +18,48 @@
...
+    @Field("asm")
     private String assemblyAccession;
 
+    @Field("tax")
     private int taxonomyAccession;
 
+    @Field("study")
     private String projectAccession;
 
     private String contig;
```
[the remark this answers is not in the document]
pabarcgar: I did this comment because I thought those fields were not being stored because they didn't have "@Field" annotation.  They are stored so this comment can be safely ignored.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/SubmittedVariantEntity.java
```diff
@@ -71,7 +96,20 @@ public SubmittedVariantEntity(Long accession, String hashedMessage, String assem
...
+        this.clusteredVariantAccession = clusteredVariantAccession;
+        this.supportedByEvidence =
+                Objects.equals(supportedByEvidence, DEFAULT_SUPPORTED_BY_EVIDENCE) ? null : supportedByEvidence;
+        this.assemblyMatch = Objects.equals(assemblyMatch, DEFAULT_ASSEMBLY_MATCH) ? null : assemblyMatch;
+        this.allelesMatch = Objects.equals(allelesMatch, DEFAULT_ALLELES_MATCH) ? null : allelesMatch;
+        this.validated = Objects.equals(validated, DEFAULT_VALIDATED) ? null : validated;
+    }
+
+    public ISubmittedVariant getModel() {
+        this.supportedByEvidence = supportedByEvidence == null ? DEFAULT_SUPPORTED_BY_EVIDENCE : supportedByEvidence;
```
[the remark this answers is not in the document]
jmmut: will do in another PR

PR #31, opened 2018-07-10
EVA-1218 Store SubSNP information in accessioning service (2)

jmmut:
It was agreed to remove the ISubmittedVariantComparator, the DTO, and put only a regular equals in SubmittedVariant.

Also, made getModel not modify the SubmittedVariantEntity class


eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariant.java
```diff
@@ -204,16 +212,72 @@ public void setValidated(Boolean validated) {
...
+        if (assemblyMatch != null ? !assemblyMatch.equals(variant.assemblyMatch) : variant.assemblyMatch != null) {
+            return false;
+        }
+        if (allelesMatch != null ? !allelesMatch.equals(variant.allelesMatch) : variant.allelesMatch != null) {
+            return false;
+        }
+        if (validated != null ? !validated.equals(variant.validated) : variant.validated != null) {
+            return false;
+        }
+        return createdDate != null ? createdDate.equals(variant.createdDate) : variant.createdDate == null;
```
pabarcgar: Were not we leaving createdDate outside the comparision?

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/SubmittedVariantEntity.java
```diff
@@ -105,11 +105,14 @@ public SubmittedVariantEntity(Long accession, String hashedMessage, String assem
     }
 
     public ISubmittedVariant getModel() {
```
jmmut: Another thing to discuss: should we allow null flags in the constructor of this entity:

allow null flags: bad because if a non-strict process doesn't compute the flags, it will be assumed the flags were computed and the result was the default value.

not allowing null flags: bad because when we query in the WS if a variant is accessioned, we will have to fill in fake flags, because the constructor will need it but they won't be used. 

thoughts @pabarcgar  and @cyenyxe ?

jmmut: agreed to add another constructor that allows not specifying the defaults, to be used for querying, but the main constructor requires them

PR #32, opened 2018-07-12
EVA-1265 Store RefSNP information in accessioning service

jmmut:
this is only the writer for RSs. the writer that will be used in the import pipeline has to combine both this and the ss one.

### DIFF
--- a/eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/ClusteredVariant.java
+++ b/eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/ClusteredVariant.java
@@ -30,35 +30,34 @@ public class ClusteredVariant implements IClusteredVariant {
 
     private long start;
 
-    private String referenceAllele;
-
-    private String alternateAllele;
+    private VariantType type;
 
     private Boolean validated;
 
     private LocalDateTime createdDate;
 
+    public ClusteredVariant(IClusteredVariant variant) {
+        this(variant.getAssemblyAccession(), variant.getTaxonomyAccession(), variant.getContig(), variant.getStart(),
+             variant.getType(), variant.isValidated());
+    }
+
     public ClusteredVariant(String assemblyAccession, int taxonomyAccession, String contig, long start,
-                            String referenceAllele, String alternateAllele, Boolean validated) {
+                            VariantType type, Boolean validated) {
         if (Objects.isNull(assemblyAccession)) {
             throw new IllegalArgumentException("Assembly accession is required");
         }
         if (Objects.isNull(contig)) {
             throw new IllegalArgumentException("Contig is required");
         }
-        if (Objects.isNull(referenceAllele)) {
-            throw new IllegalArgumentException("Reference allele is required");
-        }
-        if (Objects.isNull(alternateAllele)) {
-            throw new IllegalArgumentException("Alternate allele is required");
+        if (Objects.isNull(type)) {
+            throw new IllegalArgumentException("Variant type is required");
         }
 
         this.assemblyAccession = assemblyAccession;
         this.taxonomyAccession = taxonomyAccession;
         this.contig = contig;
         this.start = start;
-        this.referenceAllele = referenceAllele;
-        this.alternateAllele = alternateAllele;
+        this.type = type;
         this.validated = validated;
         this.createdDate = null;
     }
@@ -100,25 +99,16 @@ public class ClusteredVariant implements IClusteredVariant {
     }
 
     @Override
-    public String getReferenceAllele() {
-        return referenceAllele;
+    public VariantType getType() {
+        return type;
     }
 
-    public void setReferenceAllele(String referenceAllele) {
-        this.referenceAllele = referenceAllele;
-    }
-
-    @Override
-    public String getAlternateAllele() {
-        return alternateAllele;
-    }
-
-    public void setAlternateAllele(String alternateAllele) {
-        this.alternateAllele = alternateAllele;
+    public void setType(VariantType type) {
+        this.type = type;
     }
 
     @Override
-    public Boolean getValidated() {
+    public Boolean isValidated() {
         return validated;
     }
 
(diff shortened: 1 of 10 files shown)


eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpClusteredVariantWriterTest.java
```diff
@@ -0,0 +1,149 @@
...
+public class DbsnpClusteredVariantWriterTest {
+    private static final int TAXONOMY_1 = 3880;
+
+    private static final int TAXONOMY_2 = 3882;
+
+    private static final long EXPECTED_ACCESSION = 10000000000L;
+
+    private static final long EXPECTED_ACCESSION_2 = 10000000001L;
+
+    private static final String CONTIG_1 = "contig_1";
```
andresfsilva: Delete all unused constants

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpClusteredVariantWriterTest.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    @Test
+    public void failsOnDuplicateVariant() throws Exception {
+        ClusteredVariant clusteredVariant = new ClusteredVariant("assembly", TAXONOMY_1, "contig", START_1,
+                                                                 VARIANT_TYPE, VALIDATED);
+        DbsnpClusteredVariantEntity variant = new DbsnpClusteredVariantEntity(
+                EXPECTED_ACCESSION, hashingFunction.apply(clusteredVariant), clusteredVariant);
+
+        thrown.expect(RuntimeException.class);
```
andresfsilva: I would prefer using the `expect` in the annotation as `@Test(expected = RuntimeException.class)`. With that we know easily which test cases are expected to throw an exception

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/ISubmittedVariant.java
```diff
@@ -22,8 +22,11 @@
 import java.time.LocalDateTime;
 
 /**
- * Abstract representation of the fields that uniquely identify an accessioned submitted variant. Implemented by the
- * entity serialized into the database {@link SubmittedVariantEntity} and the message/DTO used by the REST API.
+ * Abstract representation of the fields that uniquely identify an accessioned submitted variant.
+ *
+ * Implemented by the basic bean {@link SubmittedVariant}, the entity for EVA accessions
+ * {@link SubmittedVariantEntity}, and the entities for DbSNP accessions (regular imported SSs and SSs without
```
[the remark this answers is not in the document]
jmmut: yeah, maybe "SubSnps with and without coordinates" is even better. I don't know why I thought SSs was clear at all

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpClusteredVariantWriterTest.java
jmmut: I won't open this discussion again :) there's more than one statement in the test, so I'll keep it as-is

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/summary/DbsnpClusteredVariantSummaryFunctionTest.java
```diff
@@ -0,0 +1,127 @@
...
+    private DbsnpClusteredVariantSummaryFunction summaryFunction;
+
+    @Before
+    public void setUp() {
+        summaryFunction = new DbsnpClusteredVariantSummaryFunction();
+    }
+
+    @Test
+    public void summaryFunctionMustBeIdempotent() {
+        IClusteredVariant ClusteredVariant1 =
```
cyenyxe: Variable names shouldn't be capitalized. Applies to all the methods in this class.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/persistence/DbsnpClusteredVariantEntity.java
```diff
@@ -0,0 +1,141 @@
...
+    private int taxonomyAccession;
+
+    private String contig;
+
+    private long start;
+
+    private VariantType type;
+
+    @Field("validated")
+    private Boolean validated;
```
cyenyxe: Is the annotation really necessary here?

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/ClusteredVariant.java
```diff
@@ -0,0 +1,180 @@
...
+        if (!assemblyAccession.equals(that.assemblyAccession)) {
+            return false;
+        }
+        if (!contig.equals(that.contig)) {
+            return false;
+        }
+        if (type != that.type) {
+            return false;
+        }
+        return validated != null ? validated.equals(that.validated) : that.validated == null;
```
pabarcgar: I think 'validated' should not be one of the attributes identifying the variant, as is not being used in DbsnpClusteredVariantSummaryFunction. Unless we can find two different RSs in the same taxonomy, assembly, position, type... but different validation status, but then, we should think about adding it also to the summary function.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/persistence/DbsnpClusteredVariantEntity.java
pabarcgar: Shouldn't we set the 'validated' default (let say false) to null to save space as we did with the SSs?

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/persistence/DbsnpClusteredVariantEntity.java
```diff
@@ -0,0 +1,141 @@
...
+        if (!assemblyAccession.equals(that.assemblyAccession)) {
+            return false;
+        }
+        if (!contig.equals(that.contig)) {
+            return false;
+        }
+        if (type != that.type) {
+            return false;
+        }
+        return validated != null ? validated.equals(that.validated) : that.validated == null;
```
pabarcgar: Again, validated is not in the summary function and maybe it should not be in the equals and hashCode.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpClusteredVariantWriterTest.java
```diff
@@ -0,0 +1,131 @@
...
+        assertEquals(2, accessions.size());
+        assertEquals(EXPECTED_ACCESSION, (long) accessions.get(0).getAccession());
+        assertEquals(EXPECTED_ACCESSION_2, (long) accessions.get(1).getAccession());
+
+        assertEquals(firstClusteredVariant, accessions.get(0).getModel());
+        assertEquals(secondClusteredVariant, accessions.get(1).getModel());
+    }
+
+    @Test
+    public void failsOnDuplicateVariant() throws Exception {
```
cyenyxe: Maybe also test that the flags don't have any effect in terms of considering a record duplicate?

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/persistence/DbsnpClusteredVariantEntity.java
jmmut: > we agreed to use equals as just java-wise (for tests about all the fields and such). we have the summary function to work with the (bussiness logic of) identifying fields

PR #35, opened 2018-07-18
EVA-1182 Access to legacy dbSNP SS objects with coordinates

andresfsilva:
Created Dbsnp entity, repository, database service. Modified service `SubmittedVariantAccessioningService` to use both repositories. Added tests.

### DIFF
--- a/eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningService.java
+++ b/eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningService.java
@@ -17,6 +17,14 @@
  */
 package uk.ac.ebi.eva.accession.core;
 
+import uk.ac.ebi.ampt2d.commons.accession.core.AccessionVersionsWrapper;
+import uk.ac.ebi.ampt2d.commons.accession.core.AccessionWrapper;
+import uk.ac.ebi.ampt2d.commons.accession.core.AccessioningService;
+import uk.ac.ebi.ampt2d.commons.accession.core.exceptions.AccessionCouldNotBeGeneratedException;
+import uk.ac.ebi.ampt2d.commons.accession.core.exceptions.AccessionDeprecatedException;
+import uk.ac.ebi.ampt2d.commons.accession.core.exceptions.AccessionDoesNotExistException;
+import uk.ac.ebi.ampt2d.commons.accession.core.exceptions.AccessionMergedException;
+import uk.ac.ebi.ampt2d.commons.accession.core.exceptions.HashAlreadyExistsException;
 import uk.ac.ebi.ampt2d.commons.accession.generators.monotonic.MonotonicAccessionGenerator;
 import uk.ac.ebi.ampt2d.commons.accession.hashing.SHA1HashingFunction;
 import uk.ac.ebi.ampt2d.commons.accession.persistence.jpa.monotonic.service.MonotonicDatabaseService;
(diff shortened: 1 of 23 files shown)


eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningServiceTest.java
```diff
@@ -104,7 +107,71 @@ public void sameAccessionsAreReturnedForEquivalentVariants() throws AccessionCou
...
+                                     SUPPORTED_BY_EVIDENCE, MATCHES_ASSEMBLY, ALLELES_MATCH, VALIDATED),
+                new SubmittedVariant("GCA_000009999.3", 9999, "PRJEB21999", "21", 20849999, "", "GG", CLUSTERED_VARIANT,
+                                     SUPPORTED_BY_EVIDENCE, MATCHES_ASSEMBLY, true, VALIDATED));
+
+        List<AccessionWrapper<ISubmittedVariant, String, Long>> submittedVariants = service.getOrCreate(variants);
+
+        long accession = submittedVariants.stream().filter(x -> x.getData().getProjectAccession().equals("PRJEB21794"))
+                                          .collect(Collectors.toList())
+                                          .get(0).getAccession();
+        assertEquals(5000000000L, accession);
```
cyenyxe: It would be useful to test that there is only one accession in the project.

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningServiceTest.java
```diff
@@ -104,7 +107,71 @@ public void sameAccessionsAreReturnedForEquivalentVariants() throws AccessionCou
...
+
+        long accessionDbsnp = submittedVariants.stream()
+                                               .filter(x -> x.getData().getProjectAccession().equals("PRJEB21999"))
```
cyenyxe: Avoid using a PRJEB to represent variants coming from dbSNP, because those identifiers are only issued by us.

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningServiceTest.java
```diff
@@ -104,7 +107,71 @@ public void sameAccessionsAreReturnedForEquivalentVariants() throws AccessionCou
...
+                                     SUPPORTED_BY_EVIDENCE, MATCHES_ASSEMBLY, ALLELES_MATCH, VALIDATED),
+                new SubmittedVariant("GCA_000009999.3", 9999, "PRJEB21999", "21", 20849999, "", "GG", CLUSTERED_VARIANT,
+                                     SUPPORTED_BY_EVIDENCE, MATCHES_ASSEMBLY, true, VALIDATED));
+
+        List<AccessionWrapper<ISubmittedVariant, String, Long>> accessions = service.get(variants);
+        assertEquals(2, accessions.size());
+    }
+
+    @Test
+    public void getByAccessions() {
```
cyenyxe: If you don't annotate test methods as
```java
@UsingDataSet(loadStrategy = LoadStrategyEnum.DELETE_ALL)
```
the same dataset will be used across multiple tests, creating dependencies and potentially causing bugs. See how the first methods in the class have said annotation.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningService.java
```diff
@@ -17,18 +17,114 @@
...
+        return allVariants.stream().filter(variant -> !contains(vatiantsToDelete, variant))
+                          .collect(Collectors.toList());
+    }
+
+    private Boolean contains(List<AccessionWrapper<ISubmittedVariant, String, Long>> accessionWrapperList,
+                             ISubmittedVariant iSubmittedVariant) {
+        return accessionWrapperList.stream().anyMatch(x -> x.getData().equals(iSubmittedVariant));
+    }
+
+    private List<AccessionWrapper<ISubmittedVariant, String, Long>> joinLists(List l1, List l2) {
```
jmmut: why not using generics here? if you only use this function for accession wrappers, use that type

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningService.java
```diff
@@ -17,18 +17,139 @@
...
+
+    @Override
+    public void merge(Long accession, Long accession1, String reason)
+            throws AccessionMergedException, AccessionDoesNotExistException, AccessionDeprecatedException {
+        if (accession >= accessioningMonotonicInitSs && accession1 >= accessioningMonotonicInitSs) {
+            accessioningService.merge(accession, accession1, reason);
+        } else if (accession < accessioningMonotonicInitSs && accession1 < accessioningMonotonicInitSs) {
+            accessioningServiceDbsnp.merge(accession, accession1, reason);
+        } else {
+//          TODO: Support merging submittedVariants and DbsnpSubmittedVariants
```
jmmut: Explain why it's not done. is it a design decision? is it done in a posterior commit? will it be done in another branch? If it's meant to be like this, keep only the exception

andresfsilva: According to an offline talk with @cyenyxe it will be implemented in a separate ticket [EVA-1301](https://www.ebi.ac.uk/panda/jira/browse/EVA-1301)

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/DbsnpSubmittedVariantAccessioningDatabaseService.java
```diff
@@ -0,0 +1,48 @@
...
+        super(repository,
+              accessionWrapper -> new DbsnpSubmittedVariantEntity(accessionWrapper.getAccession(),
+                                                                  accessionWrapper.getHash(),
+                                                                  accessionWrapper.getData()),
+              DbsnpSubmittedVariantEntity::getModel,
+              inactiveService);
+    }
+
+    @Override
+    public long[] getAccessionsInRanges(Collection<MonotonicRange> ranges) {
```
pabarcgar: This is related with a comment I did in other file. We are using an accessioning service in "read only" mode (well, not 100% read only because updates in the accessioned objects are possible). As a consequence of that, we have to implement the "getAccessionsInRanges" method and we don't want to generate new accessions so we have to return an empty array.

Probably create a new service that can query accessions or update accessioned objects is too much just for this. In this case, I would throw an UnsupportedOperationException when calling to this method, to make explicit that it should not be used.

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningServiceTest.java
```diff
@@ -104,7 +165,176 @@ public void sameAccessionsAreReturnedForEquivalentVariants() throws AccessionCou
...
+        assertEquals(this.submittedVariant, submittedVariant.getData());
+
+        AccessionWrapper<ISubmittedVariant, String, Long> dbsnpSubmittedVariant = service.getByAccessionAndVersion(
+                ACCESSION_DBSNP_1, 1);
+        assertEquals(this.dbsnpSubmittedVariant, dbsnpSubmittedVariant.getData());
+    }
+
+    @UsingDataSet(locations = {"/test-data/submittedVariantEntity.json"})
+    @Test
+    public void updateSubmittedVariant() throws AccessionDeprecatedException, AccessionDoesNotExistException,
```
pabarcgar: We have this test to check that we can update variants, but, can a dbSNP variant be updated? If so, another test is needed.

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningServiceTest.java
```diff
@@ -104,7 +165,176 @@ public void sameAccessionsAreReturnedForEquivalentVariants() throws AccessionCou
...
+            AccessionMergedException, HashAlreadyExistsException {
+        assertUpdate(ACCESSION, submittedVariantModified);
+    }
+
+    private void assertUpdate(long accession, ISubmittedVariant modifiedVariant)
+            throws AccessionDeprecatedException, AccessionDoesNotExistException, AccessionMergedException,
+            HashAlreadyExistsException {
+        service.update(accession, 1, modifiedVariant);
```
pabarcgar: We are doing the operation to be tested (update) and the assertion in the same method. I think the update should be moved outside this method, and maybe update the method name after that (assertVariantUpdated or something like that)

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningServiceTest.java
```diff
@@ -104,7 +165,176 @@ public void sameAccessionsAreReturnedForEquivalentVariants() throws AccessionCou
...
+    @UsingDataSet(locations = {"/test-data/submittedVariantEntity.json"})
+    @Test
+    public void patchSubmittedVariant() throws AccessionDeprecatedException, AccessionDoesNotExistException,
+            AccessionMergedException, HashAlreadyExistsException {
+        assertPatch(ACCESSION, submittedVariantModified);
+    }
+
+    private void assertPatch(long accession, ISubmittedVariant modifiedVariant) throws AccessionDeprecatedException,
+            AccessionDoesNotExistException, AccessionMergedException, HashAlreadyExistsException {
+        service.patch(accession, modifiedVariant);
```
pabarcgar: as in 'assertUpdate' method, I would move service.patch() outside this method.

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningServiceTest.java
```diff
@@ -104,7 +165,176 @@ public void sameAccessionsAreReturnedForEquivalentVariants() throws AccessionCou
...
+    @Test
+    public void deprecateSubmittedVariant() throws AccessionMergedException, AccessionDoesNotExistException,
+            AccessionDeprecatedException {
+        assertDeprecate(ACCESSION_TO_DEPRECATE, "Test deprecate");
+    }
+
+    private void assertDeprecate(long accession, String reason) throws AccessionMergedException,
+            AccessionDoesNotExistException, AccessionDeprecatedException {
+        assertFalse(service.getByAccessions(Collections.singletonList(accession)).isEmpty());
+        service.deprecate(accession, reason);
```
pabarcgar: Same comment that in the other 'assert' methods

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningServiceTest.java
```diff
@@ -104,7 +165,176 @@ public void sameAccessionsAreReturnedForEquivalentVariants() throws AccessionCou
...
+    public void mergeSubmittedVariant() throws AccessionMergedException, AccessionDoesNotExistException,
+            AccessionDeprecatedException {
+        assertMerge(ACCESSION_TO_MERGE_1, ACCESSION_TO_MERGE_2, "Test merge");
+    }
+
+    private void assertMerge(long accessionOrigin, long accessionDestination, String reason) throws
+            AccessionMergedException, AccessionDoesNotExistException, AccessionDeprecatedException {
+        assertFalse(service.getByAccessions(Collections.singletonList(accessionOrigin)).isEmpty());
+        assertFalse(service.getByAccessions(Collections.singletonList(accessionDestination)).isEmpty());
+        service.merge(accessionOrigin, accessionDestination, reason);
```
pabarcgar: Same comment that in other 'assert' methods

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningServiceTest.java
andresfsilva: There is a separate test called `updateDbsnpSubmittedVariant`

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/DbsnpSubmittedVariantAccessioningDatabaseService.java
jmmut: agree on throwing unsuported operations wherever applies

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/persistence/DbsnpMonotonicAccessionGenerator.java
```diff
@@ -0,0 +1,56 @@
...
+public class DbsnpMonotonicAccessionGenerator<T> extends MonotonicAccessionGenerator {
+
+    public DbsnpMonotonicAccessionGenerator(long blockSize, String categoryId, String applicationInstanceId,
+                                            ContiguousIdBlockService contiguousIdBlockService) {
+        super(blockSize, categoryId, applicationInstanceId, contiguousIdBlockService);
+    }
+
+    @Override
+    public synchronized void recoverState(long[] committedElements) throws AccessionIsNotPendingException {
+        throw new UnsupportedOperationException("New accessions should not be generated");
```
pabarcgar: This message is too generic. I would add a reference that dbSNP accessions are the ones that we cannot generate. 

"New accessions cannot be issued for dbSNP variants"

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningServiceTest.java
pabarcgar: 👍

PR #36, opened 2018-07-19
EVA-1282 Map RS read from dbSNP database to Java model

jmmut:
this PR was made on top of #33  and #34 , so merge those first.

processes present as of this PR: 
- [x] 9 forward strand
- [x] 5 alleles match (ref is included in alleles list)
- [x] 2 assembly check
- [x] 3 variant instantiation (history operations still not done)
  - [x] ss instantiation
  - [x] rs instantiation
  - [ ] history operations (won't be done if we use the service directly)
- [x] 1 classification (still needs something of the declustering)
- [ ] 4 normalize and 4.1 renormalize 
  - [x] trim "-" and spaces
  - [ ] normalize multiallelic
  - [ ] renormalize context base
- [ ] 7 store rs
- [ ] 6 store ss
- [ ] 8 decluster

PR #37, opened 2018-07-26
EVA-1277 Decluster SS from RS when REF allele is not in the list of alleles

jmmut:
this PR was made on top of #36 and https://github.com/EBIvariation/accession-commons/pull/42 , so merge those first and deploy accession-commons/0.5-SNAPSHOT.

processes present as of this PR: 
- [x] 9 forward strand
- [x] 5 alleles match (ref is included in alleles list)
- [x] 2 assembly check
- [x] 3 variant instantiation (history operations still not done)
  - [x] ss instantiation
  - [x] rs instantiation
  - [ ] history operations (won't be done if we use the service directly)
- [x] 1 classification (still needs something of the declustering)
- [ ] 4 normalize and 4.1 renormalize 
  - [x] trim "-" and spaces
  - [ ] normalize multiallelic (we didn't do this for the main warehouse import, I think)
  - [ ] renormalize context base
- [x] 7 store rs
  - [x] don't stop the step when the same RS is inserted several times (*)
- [x] 6 store ss
- [ ] 8 decluster
  - [x] decluster if alleles mismatch
  - [ ] decluster if types differ (within multiallelic and across ss)

(*) as we process each dbsnp row independently, and several rows can refer to the same RS, we said we would do several inserts in the RS table for the same RS, but we have to catch the duplicate key exception and ignore it. This doesn't apply to SS insertion, as for EVA-1188 we want to merge two SS that have the same identifying fields (same hash).


eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessor.java
```diff
@@ -16,26 +16,100 @@
...
         return dbsnpVariantsWrapper;
     }
+
+    private SubmittedVariant subSnpNoHgvsToSubmittedVariant(SubSnpNoHgvs subSnpNoHgvs, String alternate) {
+        Region variantRegion = getVariantRegion(subSnpNoHgvs);
+        String reference = subSnpNoHgvs.getReferenceInForwardStrand();
+        SubmittedVariant variant = new SubmittedVariant(subSnpNoHgvs.getAssembly(), subSnpNoHgvs.getTaxonomyId(),
+                                                        getProjectAccession(subSnpNoHgvs),
+                                                        variantRegion.getChromosome(), variantRegion.getStart(),
+                                                        reference, alternate, subSnpNoHgvs.getRsId());
```
cyenyxe: Calling the constructor without flags will set their default values, even though at this point we don't know if it matches the assembly or not. I think it would be better to use the constructor will all the arguments, and setting the assembly match as null.

We could even reorder the constructor parameters to something like supported by evidence, validated, alleles match and assembly match. This will place the null at the very end and kind of replicate the application flow, but could cause issues in the tests if the refactoring doesn't reorder the input automatically.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriterTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+    private static final Long SUBMITTED_VARIANT = 15L;
+
+    private static final Boolean SUPPORTED_BY_EVIDENCE = true;
+
+    private static final Boolean MATCHES_ASSEMBLY = false;
+
+    private static final Boolean ALLELES_MATCH = true;
+
+    private static final Boolean VALIDATED = false;
```
cyenyxe: If using the defaults would make no difference, it could be nicer to just use those from `ISubmittedVariant`.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriterTest.java
```diff
@@ -0,0 +1,245 @@
...
+        SubmittedVariantAccessioningConfiguration.class})
+public class DbsnpVariantsWriterTest {
+
+    private static final int TAXONOMY_1 = 3880;
+
+    private static final int TAXONOMY_2 = 3882;
+
+    private static final long EXPECTED_ACCESSION = 10000000000L;
+
+    private static final int START_1 = 100;
```
cyenyxe: There is only one start, so the suffix shouldn't be necessary.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriterTest.java
```diff
@@ -0,0 +1,245 @@
...
+    private void assertRsWasStored(DbsnpVariantsWrapper wrapper) {
+        List<DbsnpClusteredVariantEntity> rsEntities = mongoTemplate.find(new Query(),
+                                                                          DbsnpClusteredVariantEntity.class);
+        assertEquals(1, rsEntities.size());
+        assertEquals(wrapper.getClusteredVariant(), rsEntities.get(0));
+    }
+
+    @Test
+    public void writeComplexVariant() throws Exception {
+        SubmittedVariant submittedVariant_1 = new SubmittedVariant("assembly", TAXONOMY_1, "project", "contig", START_1,
```
cyenyxe: Our naming rule is `submittedVariant1` and `submittedVariant2`.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriterTest.java
```diff
@@ -0,0 +1,245 @@
...
+                                                                 SUPPORTED_BY_EVIDENCE, MATCHES_ASSEMBLY, ALLELES_MATCH,
+                                                                 VALIDATED);
+        DbsnpVariantsWrapper wrapper = buildSimpleWrapper();
+        wrapper.setSubmittedVariants(Arrays.asList(
+                new DbsnpSubmittedVariantEntity(SUBMITTED_VARIANT,
+                                                hashingFunctionSubmitted.apply(submittedVariant_1),
+                                                submittedVariant_1),
+                new DbsnpSubmittedVariantEntity(SUBMITTED_VARIANT,
+                                                hashingFunctionSubmitted.apply(submittedVariant_2),
+                                                submittedVariant_2)));
```
cyenyxe: This is not a clean way to constructor the test object, because it sets with certain values and then modifies it. It gets even worse in `repeatedClusteredVariants` because it initialized multiple objects in different places.

If needed, create another private method that receives the submitted variants you want to initialize the wrapper with.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriterTest.java
```diff
@@ -0,0 +1,245 @@
...
+        DbsnpVariantsWrapper wrapper = buildSimpleWrapper();
+
+        dbsnpVariantsWriter.write(Collections.singletonList(wrapper));
+
+        List<DbsnpSubmittedVariantEntity> ssEntities = mongoTemplate.find(new Query(),
+                                                                          DbsnpSubmittedVariantEntity.class);
+        assertEquals(1, ssEntities.size());
+        assertEquals(wrapper.getSubmittedVariants().get(0), ssEntities.get(0));
+
+        assertRsWasStored(wrapper);
```
cyenyxe: The assert style is inconsistent: submitted variants are checked in the test method, but the clustered variant is checked delegating into another. If needed created another method that receives the number of expected SS, and that will allow you to loop through the retrieved entities in a generic way.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessorTest.java
```diff
@@ -1,35 +1,23 @@
-/*
```
cyenyxe: Restore the copyright header (although it should be dated 2018 only).

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/configuration/ImportDbsnpVariantsWriterConfiguration.java
```diff
@@ -34,7 +35,9 @@
 
     @Bean(name = DBSNP_VARIANT_WRITER)
     @StepScope
-    DbsnpVariantsWriter dbsnpVariantWriter(InputParameters parameters, MongoTemplate mongoTemplate) throws Exception {
+    DbsnpVariantsWriter dbsnpVariantWriter(InputParameters parameters,
+                                           MongoTemplate mongoTemplate,
+                                           SubmittedVariantAccessioningService service) throws Exception {
```
pabarcgar: I think this service is not being used.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
```diff
@@ -15,25 +15,69 @@
...
+        }
+        writeClusteredVariants(wrappers);
+    }
+
+    private void writeClusteredVariants(List<? extends DbsnpVariantsWrapper> items) {
+        try {
+            Collection<DbsnpClusteredVariantEntity> uniqueClusteredVariants =
+                    items.stream()
+                         .map(DbsnpVariantsWrapper::getClusteredVariant)
+                         .collect(Collectors.toMap(DbsnpClusteredVariantEntity::getHashedMessage,
```
pabarcgar: couldn't this be done with a simple map()? (because we are just getting the values in the next operation)

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToClusteredVariantProcessor.java
```diff
@@ -39,20 +39,27 @@ public SubSnpNoHgvsToClusteredVariantProcessor() {
         hashingFunction = new DbsnpClusteredVariantSummaryFunction().andThen(new SHA1HashingFunction());
     }
 
+    /**
+     * Instantiate a {@link DbsnpClusteredVariantEntity} from a {@link SubSnpNoHgvs}.
+     *
+     * We define the type of the ClusteredVariant (RefSnp) as the type of the first alternate allele (according to
```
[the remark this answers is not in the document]
andresfsilva: According to the offline discussion all types will be directly mapped to it's corresponding type except for `DIV`.

When the dbsnp type is `DIV` the type will be determined by the first SS which has as type `INDEL`, `INS` or `DEL` according to the `VariantClassifier`.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessorTest.java
```diff
@@ -51,13 +39,41 @@
...
+        assertEquals(1, dbsnpVariantsWrapper.getSubmittedVariants().size());
+        assertNull(dbsnpVariantsWrapper.getSubmittedVariants().get(0).getClusteredVariantAccession());
+    }
+
+    private void assertDeclusterOperationsRegistered(SubSnpNoHgvs subSnpNoHgvs,
+                                                     DbsnpVariantsWrapper dbsnpVariantsWrapper) {
+        assertEquals(1, dbsnpVariantsWrapper.getOperations().size());
+        assertEquals(EventType.UPDATED, dbsnpVariantsWrapper.getOperations().get(0).getEventType());
+        assertEquals(1, dbsnpVariantsWrapper.getOperations().get(0).getInactiveObjects().size());
+        assertEquals("Declustered (Alleles mismatch)", dbsnpVariantsWrapper.getOperations().get(0).getReason());
```
pabarcgar: If we add other decluster operations to this processor in the future, those method names should be improved (or add some parameter with the decluster reason)

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessor.java
andresfsilva: `assemblyMatch` can't be null in `SubmittedVariantEntity` so it will be set to `false` but the real value will need to be determined by the assembly checker in a later step

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToClusteredVariantProcessor.java
```diff
@@ -72,4 +80,48 @@ public DbsnpClusteredVariantEntity process(SubSnpNoHgvs subSnpNoHgvs) throws Exc
...
 
+    /**
+     * If the RefSnp type cannot be determined because none of the variants are type INS, DEL or INDEL, the type INDEL
+     * is assigned by default and the SubSnps would be declustered in
+     * {@link SubSnpNoHgvsToDbsnpVariantsWrapperProcessor }
+     */
+    private VariantType getVariantTypeWhenDiv(SubSnpNoHgvs subSnpNoHgvs) {
+        VariantType variantType = null;
+        List<String> alternateAlleles = subSnpNoHgvs.getAlternateAllelesInForwardStrand();
+        VariantType ssVariantType;
```
cyenyxe: It is not needed to track the value of this outside the loop, so its declaration can be moved inside it.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToClusteredVariantProcessor.java
```diff
@@ -72,4 +80,48 @@ public DbsnpClusteredVariantEntity process(SubSnpNoHgvs subSnpNoHgvs) throws Exc
         return variantEntity;
     }
 
+    /**
+     * If the RefSnp type cannot be determined because none of the variants are type INS, DEL or INDEL, the type INDEL
+     * is assigned by default and the SubSnps would be declustered in
+     * {@link SubSnpNoHgvsToDbsnpVariantsWrapperProcessor }
```
cyenyxe: Suggested rewording: 

> If the RefSnp type cannot be determined because none of the associated SubSnps are type INS, DEL or INDEL, the type INDEL is assigned by default, and the SubSnps should be declustered by {@link SubSnpNoHgvsToDbsnpVariantsWrapperProcessor}.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessor.java
```diff
@@ -48,68 +53,78 @@ public SubSnpNoHgvsToDbsnpVariantsWrapperProcessor() {
...
         List<DbsnpSubmittedVariantOperationEntity> operations = new ArrayList<>();
         DbsnpClusteredVariantEntity clusteredVariant = subSnpNoHgvsToClusteredVariantProcessor.process(subSnpNoHgvs);
 
         List<String> alternateAlleles = subSnpNoHgvs.getAlternateAllelesInForwardStrand();
         for (String alternateAllele : alternateAlleles) {
             SubmittedVariant variant = subSnpNoHgvsToSubmittedVariant(subSnpNoHgvs, alternateAllele);
             if (!variant.isAllelesMatch()) {
-                decluster(subSnpNoHgvs.getSsId(), variant, operations);
+                decluster(subSnpNoHgvs.getSsId(), variant, operations,
+                          "None of the variant alleles match the reference allele");
```
cyenyxe: I think it is important that we still state the variant was declustered, like you did before:

> Declustered: None of the variant alleles match the reference allele

PR #41, opened 2018-08-01
EVA-1302 Renormalize ambioguous alleles in accession service import

pabarcgar:


eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/RenormalizationProcessorTest.java
```diff
@@ -0,0 +1,168 @@
...
+        DbsnpSubmittedVariantEntity variant2 = new DbsnpSubmittedVariantEntity(anotherSsId, anotherHash, ASSEMBLY,
+                                                                               TAXONOMY, otherProject, "22", 6, "C", "",
+                                                                               anotherRsId, !IS_SUPPORTED_BY_EVIDENCE,
+                                                                               !MATCHES_ASSEMBLY, !ALLELES_MATCH,
+                                                                               !VALIDATED, 3);
+
+        List<DbsnpSubmittedVariantEntity> normalizedVariants = renormalizer.process(Arrays.asList(variant, variant2));
+
+
+        // the version is compared in the equals method, so we have to compare it explicitly
```
cyenyxe: And created date too? It should be set as the value from the dbSNP database.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/RenormalizationProcessor.java
```diff
@@ -0,0 +1,141 @@
...
+
+import uk.ac.ebi.eva.accession.core.ISubmittedVariant;
+import uk.ac.ebi.eva.accession.core.io.FastaSequenceReader;
+import uk.ac.ebi.eva.accession.core.persistence.DbsnpSubmittedVariantEntity;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class RenormalizationProcessor implements
+        ItemProcessor<List<DbsnpSubmittedVariantEntity>, List<DbsnpSubmittedVariantEntity>> {
```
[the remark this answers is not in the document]
pabarcgar: I tried that first, but then I needed DbsnpSubmittedVariantEntity objects (those are the ones that have accession, hash, and version). 

I can make a casting inside one of the private methods and make the public signatures depend on ISubmittedVariant. Do you prefer that way?

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/RenormalizationProcessorTest.java
cyenyxe: Variants are manually initialized instead of read from any source, so the created date is not initialized.

PR #42, opened 2018-08-03
EVA-1281 Decluster when SS and RS types differ

andresfsilva:
Decluster when SS and RS types differ

### DIFF
--- a/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessor.java
+++ b/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessor.java
@@ -45,11 +45,13 @@ import static uk.ac.ebi.eva.accession.core.ISubmittedVariant.DEFAULT_VALIDATED;
 
 public class SubSnpNoHgvsToDbsnpVariantsWrapperProcessor implements ItemProcessor<SubSnpNoHgvs, DbsnpVariantsWrapper> {
 
-    private static final String DECLUSTERED_ALLELES_AND_TYPE_MISMATCH = "Declustered (Alleles and type mismatch)";
+    static final String DECLUSTERED = "Declustered: ";
 
-    private static final String DECLUSTERED_ALLELES_MISMATCH = "Declustered (Alleles mismatch)";
+    static final String DECLUSTERED_ALLELES_MISMATCH =
+            "None of the variant alleles match the reference allele. ";
 
-    private static final String DECLUSTERED_TYPE_MISMATCH = "Declustered (Type mismatch)";
+    static final String DECLUSTERED_TYPE_MISMATCH =
+            "The variant type inferred from the alleles does not match the one asserted by dbSNP. ";
 
     private SubmittedVariantRenormalizationProcessor renormalizationProcessor;
 
@@ -73,14 +75,27 @@ public class SubSnpNoHgvsToDbsnpVariantsWrapperProcessor implements ItemProcesso
         List<String> alternateAlleles = subSnpNoHgvs.getAlternateAllelesInForwardStrand();
         for (String alternateAllele : alternateAlleles) {
             SubmittedVariant submittedVariant = subSnpNoHgvsToSubmittedVariant(subSnpNoHgvs, alternateAllele);
-            boolean typeMatches = isSameType(clusteredVariant, submittedVariant, subSnpNoHgvs.getDbsnpVariantType());
-            if (!submittedVariant.isAllelesMatch() && !typeMatches) {
-                decluster(subSnpNoHgvs.getSsId(), submittedVariant, operations, DECLUSTERED_ALLELES_AND_TYPE_MISMATCH);
-            } else if (!submittedVariant.isAllelesMatch()) {
-                decluster(subSnpNoHgvs.getSsId(), submittedVariant, operations, DECLUSTERED_ALLELES_MISMATCH);
-            } else if (!typeMatches) {
-                decluster(subSnpNoHgvs.getSsId(), submittedVariant, operations, DECLUSTERED_TYPE_MISMATCH);
+
+            List<String> reasons = new ArrayList<>();
+            if (!submittedVariant.isAllelesMatch()) {
+                reasons.add(DECLUSTERED_ALLELES_MISMATCH);
             }
+            boolean typeMatches = isSameType(clusteredVariant, submittedVariant, subSnpNoHgvs.getDbsnpVariantType());
+            if (!typeMatches) {
+                reasons.add(DECLUSTERED_TYPE_MISMATCH);
+            }
+
+            if (!reasons.isEmpty()) {
+                decluster(subSnpNoHgvs.getSsId(), submittedVariant, operations, reasons);
+            }
+
+//            if (!submittedVariant.isAllelesMatch() && !typeMatches) {
+//                decluster(subSnpNoHgvs.getSsId(), submittedVariant, operations, DECLUSTERED_ALLELES_AND_TYPE_MISMATCH);
+//            } else if (!submittedVariant.isAllelesMatch()) {
+//                decluster(subSnpNoHgvs.getSsId(), submittedVariant, operations, DECLUSTERED_ALLELES_MISMATCH);
+//            } else if (!typeMatches) {
+//                decluster(subSnpNoHgvs.getSsId(), submittedVariant, operations, DECLUSTERED_TYPE_MISMATCH);
+//            }
 
             String hash = hashingFunction.apply(submittedVariant);
             DbsnpSubmittedVariantEntity submittedVariantEntity = new DbsnpSubmittedVariantEntity(subSnpNoHgvs.getSsId(),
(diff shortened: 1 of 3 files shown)


eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessor.java
```diff
@@ -86,6 +100,13 @@ public DbsnpVariantsWrapper process(SubSnpNoHgvs subSnpNoHgvs) throws Exception
         return dbsnpVariantsWrapper;
     }
 
+    private boolean isSameType(DbsnpClusteredVariantEntity clusteredVariant, SubmittedVariant submittedVariant, DbsnpVariantType dbsnpVariantType) {
```
cyenyxe: This line is longer than 120, please split it.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessor.java
```diff
@@ -43,8 +45,16 @@
 
 public class SubSnpNoHgvsToDbsnpVariantsWrapperProcessor implements ItemProcessor<SubSnpNoHgvs, DbsnpVariantsWrapper> {
 
+    private static final String DECLUSTERED_ALLELES_AND_TYPE_MISMATCH = "Declustered (Alleles and type mismatch)";
+
+    private static final String DECLUSTERED_ALLELES_MISMATCH = "Declustered (Alleles mismatch)";
+
+    private static final String DECLUSTERED_TYPE_MISMATCH = "Declustered (Type mismatch)";
```
cyenyxe: We had defined a nicer message in the previous PR, so these would be:

Alleles mismatch: `None of the variant alleles match the reference allele`
Type mismatch: `The variant type inferred from the alleles does not match the one asserted by dbSNP`

More comments on how to construct the message below.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessor.java
```diff
@@ -43,8 +45,16 @@
 
 public class SubSnpNoHgvsToDbsnpVariantsWrapperProcessor implements ItemProcessor<SubSnpNoHgvs, DbsnpVariantsWrapper> {
 
+    private static final String DECLUSTERED_ALLELES_AND_TYPE_MISMATCH = "Declustered (Alleles and type mismatch)";
+
+    private static final String DECLUSTERED_ALLELES_MISMATCH = "Declustered (Alleles mismatch)";
```
pabarcgar: I think that "Declustered: None of the variant alleles match the reference allele" is more informative (as was suggested in other PR). If this message is updated, the other messages should be updated too.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessor.java
```diff
@@ -63,9 +73,13 @@ public DbsnpVariantsWrapper process(SubSnpNoHgvs subSnpNoHgvs) throws Exception
...
-            if (!submittedVariant.isAllelesMatch()) {
-                decluster(subSnpNoHgvs.getSsId(), submittedVariant, operations,
-                          "Declustered: None of the variant alleles match the reference allele");
+            boolean typeMatches = isSameType(clusteredVariant, submittedVariant, subSnpNoHgvs.getDbsnpVariantType());
+            if (!submittedVariant.isAllelesMatch() && !typeMatches) {
+                decluster(subSnpNoHgvs.getSsId(), submittedVariant, operations, DECLUSTERED_ALLELES_AND_TYPE_MISMATCH);
+            } else if (!submittedVariant.isAllelesMatch()) {
+                decluster(subSnpNoHgvs.getSsId(), submittedVariant, operations, DECLUSTERED_ALLELES_MISMATCH);
+            } else if (!typeMatches) {
+                decluster(subSnpNoHgvs.getSsId(), submittedVariant, operations, DECLUSTERED_TYPE_MISMATCH);
```
cyenyxe: These combinatorial combinations could become problematic if we find more reasons to decluster. What about providing a list of reasons as an argument, then the message could be built prefixing them with a single `Declustered` and concatenating them all?

The most complex output message could look something like this:

> Declustered: None of the variant alleles match the reference allele. The variant type inferred from the alleles does not match the one asserted by dbSNP.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessorTest.java
```diff
@@ -45,6 +60,12 @@
 
     private static SubSnpNoHgvsToDbsnpVariantsWrapperProcessor processor;
 
+    private static final String DECLUSTERED_ALLELES_AND_TYPE_MISMATCH = "Declustered (Alleles and type mismatch)";
+
+    private static final String DECLUSTERED_ALLELES_MISMATCH = "Declustered (Alleles mismatch)";
+
+    private static final String DECLUSTERED_TYPE_MISMATCH = "Declustered (Type mismatch)";
+
```
cyenyxe: If you are going to check that the constants match, make them public and compare against them directly. Otherwise just a single character change will break the tests.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessorTest.java
```diff
@@ -60,31 +81,87 @@ public static void tearDownClass() throws Exception {
...
+        //Check Rs
         assertEquals(subSnpNoHgvs.getRsId(), dbsnpVariantsWrapper.getClusteredVariant().getAccession());
     }
 
-    private void assertSubmittedVariantDeclustered(DbsnpVariantsWrapper dbsnpVariantsWrapper) {
-        assertEquals(1, dbsnpVariantsWrapper.getSubmittedVariants().size());
+    @Test
+    public void processSubmittedVariantTypeMismatch() throws Exception {
+        SubSnpNoHgvs subSnpNoHgvsTypeMismatch = new SubSnpNoHgvs(25928972L, 14718243L, "G", "G/-/T/TT", ASSEMBLY,
```
pabarcgar: Looking at the alleles, I see a SNP (G -> T), a deletion (G -> -) and an insertion (G -> TT). The dbsnp type is DIV, so both the insertion and deletion will be under that type. Why is one of those declustered? is the RS type the one of the first found (deletion or insertion) and then the other is failing?

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessorTest.java
```diff
@@ -60,31 +81,87 @@ public static void tearDownClass() throws Exception {
...
+                                                                 Orientation.FORWARD, false, false, false, false,
+                                                                 CREATED_DATE, CREATED_DATE, TAXONOMY);
+
+        DbsnpVariantsWrapper dbsnpVariantsWrapper = processor.process(subSnpNoHgvsTypeMismatch);
+
+        //Check Decluster
+        assertEquals(3, dbsnpVariantsWrapper.getSubmittedVariants().size());
+        assertNotNull(dbsnpVariantsWrapper.getSubmittedVariants().get(0).getClusteredVariantAccession());
+        assertNull(dbsnpVariantsWrapper.getSubmittedVariants().get(1).getClusteredVariantAccession());
+        assertNull(dbsnpVariantsWrapper.getSubmittedVariants().get(2).getClusteredVariantAccession());
```
cyenyxe: I don't understand which is the order of the alternate alleles here but as @pabarcgar said, it doesn't look like they should be declustered this way.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessorTest.java
andresfsilva: The dbsnpType is DIV so the EVA type of the RS is determined in `DbsnpClusteredVariantEntity` using the method `getVariantTypeWhenDiv` in this case and the result would be `DEL`. 

This is because when the dbsnpType is DIV we will determine the RS EVA type according to the type of the first variant which is INS, DEL or INDEL type. 

In this test the RS type would be DEL so the variants with type SNP and INS would be declustered for type mismatch.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessorTest.java
andresfsilva: G is the reference so it is not counted as variant, so we have -/T/TT as alternates. G>'' (DEL) would be the first one and this will determine the RS EVA type. The other variants will be declustered (SNV, INS)

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessorTest.java
cyenyxe: This answer addresses on of my comments too.

PR #48, opened 2018-08-15
EVA-1221 Decluster processor refactor

andresfsilva:


eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubmittedVariantDeclusterProcessor.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public DbsnpVariantsWrapper process(DbsnpVariantsWrapper dbsnpVariantsWrapper) throws Exception {
+        List<DbsnpSubmittedVariantEntity> submittedVariants = new ArrayList<>();
+        List<DbsnpSubmittedVariantOperationEntity> operations = new ArrayList<>();
+
+        for (DbsnpSubmittedVariantEntity submittedVariantEntity: dbsnpVariantsWrapper.getSubmittedVariants()) {
+            SubmittedVariant submittedVariant = new SubmittedVariant(submittedVariantEntity);
+            List<String> reasons = new ArrayList<>();
+            if (variantNeedsDecluster(dbsnpVariantsWrapper, submittedVariant, reasons)) {
+                decluster(submittedVariantEntity, submittedVariant, operations, submittedVariants, reasons);
```
cyenyxe: The entity created in this method could be returned instead of added to the `submittedVariants` collection.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubmittedVariantDeclusterProcessor.java
```diff
@@ -0,0 +1,114 @@
...
+        StringBuilder reason = new StringBuilder(DECLUSTERED);
+        reasons.forEach(reason::append);
+        Long accession = nonDeclusteredVariantEntity.getAccession();
+        operation.fill(EventType.UPDATED, accession, null, reason.toString(),
+                       Collections.singletonList(inactiveEntity));
+        operations.add(operation);
+
+        //Decluster submitted variant
+        variant.setClusteredVariantAccession(null);
+        String hash = nonDeclusteredVariantEntity.getHashedMessage();
```
cyenyxe: Agreed offline to recalculate hash just in case the summary function is modified in the future to include the RS accession.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessorTest.java
```diff
@@ -79,7 +73,7 @@ public static void tearDownClass() throws Exception {
     }
 
     @Test
-    public void processSubmittedVariantAllelesMismatch() throws Exception {
+    public void processSubSnpNoHgvsToWrapper() throws Exception {
```
cyenyxe: For the sake of completion, could you implement another test with multiple submitted variants?

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubmittedVariantDeclusterProcessorTest.java
```diff
@@ -0,0 +1,192 @@
...
+        assertEquals(1, wrapper.getSubmittedVariants().size());
+        assertNull(wrapper.getSubmittedVariants().get(0).getClusteredVariantAccession());
+
+        //Check Operations
+        assertOperation(wrapper, getReason(Collections.singletonList(DECLUSTERED_ALLELES_MISMATCH)),
+                        wrapper.getSubmittedVariants().get(0).getAccession(), 0);
+    }
+
+    @Test
+    public void processSubmittedVariantTypeMismatch() throws Exception {
```
cyenyxe: The input for this method is the same as in `processSubmittedVariantAllelesAndTypeMismatch`. Modify this one so it only checks the case of type mismatch.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubmittedVariantDeclusterProcessorTest.java
```diff
@@ -0,0 +1,192 @@
...
+        assertOperation(wrapper, getReason(Arrays.asList(DECLUSTERED_ALLELES_MISMATCH, DECLUSTERED_TYPE_MISMATCH)),
+                        wrapper.getSubmittedVariants().get(1).getAccession(), 1);
+        assertOperation(wrapper, getReason(Arrays.asList(DECLUSTERED_ALLELES_MISMATCH, DECLUSTERED_TYPE_MISMATCH)),
+                        wrapper.getSubmittedVariants().get(2).getAccession(), 2);
+    }
+
+    private List<DbsnpSubmittedVariantEntity> createSubmittedVariantEntities(String reference, String alternate,
+                                                                             boolean allelesMatch) {
+        List<DbsnpSubmittedVariantEntity> submittedVariantEntities = new ArrayList<>();
+        String[] alternates = Arrays.stream(StringUtils.split(alternate, "/")).toArray(String[]::new);
```
cyenyxe: It would be easier to make `alternate` into an array argument instead of a string, as in `new String[]{ "-", "T", "TT" }`. It reduces a bit the copy & paste possibilities, but there aren't many of these test cases anyway.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubmittedVariantDeclusterProcessorTest.java
andresfsilva: The difference in the input is the `allelesMatch` flag. In this case the value is `true` so it will not decluster for alleles mismatch, only in case of type mismatch. 

The result of this test is 2 variants declustered out of 3, while the `processSubmittedVariantAllelesAndTypeMismatch` has all the variants declustered.

cyenyxe: I think it is very counter-intuitive to build an object with alleles that mismatch, but the flag is set to express the opposite...

andresfsilva: The input in this processor is the wrapper and it doesn't have a variant with the same reference and alternate, it only has the 'allelesMatch` flag on every variant. If we add the reference in the alternates list it will create a variant that shouldn't be in the wrapper.  

`SubSnpNoHgvsToDbsnpVariantsWrapperProcessor` is called before the decluster and it calculates `allelesMatch` flag and create one variant for each alternate allele except when the alternate is the same reference.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubmittedVariantDeclusterProcessor.java
```diff
@@ -87,28 +91,26 @@ private boolean isSameType(DbsnpClusteredVariantEntity clusteredVariant, Submitt
                                 .equals(clusteredVariant.getType());
```
cyenyxe: Could you wrap this call with a try - catch of `IllegalArgumentException` and return false in that case? We agreed offline to consider that a type mismatch.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubmittedVariantDeclusterProcessor.java
```diff
@@ -0,0 +1,120 @@
...
+
+    }
+
+    @Override
+    public DbsnpVariantsWrapper process(DbsnpVariantsWrapper dbsnpVariantsWrapper) throws Exception {
+        List<DbsnpSubmittedVariantEntity> submittedVariants = new ArrayList<>();
+        List<DbsnpSubmittedVariantOperationEntity> operations = new ArrayList<>();
+
+        for (DbsnpSubmittedVariantEntity submittedVariantEntity: dbsnpVariantsWrapper.getSubmittedVariants()) {
+            SubmittedVariant submittedVariant = new SubmittedVariant(submittedVariantEntity);
```
pabarcgar: why do we need to extract a SubmittedVariant object?

PR #57, opened 2018-08-30
EVA-1287 Pilot for import into accessioning service - corner cases (use eva study id instead of dbsnp's)

jmmut:
Some dbsnp batches have also a EVA study id (PRJ*) because they were submitted to both places or brokered. For those we prefer using the EVA study. 

This PR allows having an extra table named `projectAccessionMapping` with the structure of `ProjectAccessionMapping`, exemplified in eva-accession-import/src/test/resources/test-data/projectAccessionMapping.json, with documents like this:
```
    {
      "evaStudyId": "PRJEB23",
      "dbsnpBatchHandle": "handle_1",
      "dbsnpBatchName": "batch_name_1",
      "taxonomyAccession": 42
    }
```
Remember to insert as NumberInt if the mapping is stored manually using the mongo shell.


eva-accession-import/pom.xml
```diff
@@ -40,6 +40,12 @@
             <artifactId>postgresql</artifactId>
             <scope>runtime</scope>
         </dependency>
+        <dependency>
+            <groupId>com.lordofthejars</groupId>
+            <artifactId>nosqlunit-mongodb</artifactId>
+            <version>1.0.0-rc.5</version>
+            <scope>test</scope>
+        </dependency>
```
cyenyxe: I have noticed this dependency is already declared in `eva-accession-core`. It should be moved to the parent POM, then declared (without version) in the children.

eva-accession-import/src/test/resources/application.properties
```diff
@@ -14,4 +14,9 @@ parameters.chunkSize=100
 parameters.forceRestart=false
 parameters.pageSize=100
 
+spring.data.mongodb.database=dbsnp-import-test
+spring.data.mongodb.host=localhost
+spring.data.mongodb.password=
+spring.data.mongodb.port=27017
```
cyenyxe: I think it would be preferable to move these Mongo properties to a separate properties file because:
* They are specific to a single test
* Said test doesn't need the rest of the properties

The database could then be renamed like that test, to avoid clashes if/when others decide to use nosqlunit too.

jmmut: the other points are valid, but that test does need the properties for the datasource and the parameters to be able to load the spring application context to use the NoSqlUnit and mongoTemplate.

PR #60, opened 2018-09-03
EVA-1287 Pilot for import into accessioning service - corner cases (assembly check and assembly report)

jmmut:
this PR improves the assembly checks and adds the column "assigned molecule" from the assembly report if there are no duplicates. 

The assigned molecule is necessary for species such as [carrot](ftp://ftp.ncbi.nlm.nih.gov/genomes/all/GCA/001/625/215/GCA_001625215.1_ASM162521v1/GCA_001625215.1_ASM162521v1_assembly_report.txt) where "assigned molecule" is the only column where you can search using the simple chromosome names such as "1". We wouldn't be able to make any assembly check on those species otherwise.


eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/contig/ContigMapping.java
```diff
@@ -20,50 +20,98 @@
...
+        genBankToSynonyms.forEach((key, contigSynonym) -> contigSynonym.setAssignedMolecule(null));
+        refSeqToSynonyms.forEach((key, contigSynonym) -> contigSynonym.setAssignedMolecule(null));
+    }
+
     private void populateMaps(AssemblyReportReader assemblyReportReader) throws Exception {
         ContigSynonyms contigSynonyms;
         while ((contigSynonyms = assemblyReportReader.read()) != null) {
             fillContigConventionMaps(contigSynonyms);
         }
+        if (assignedMoleculeHasDuplicates) {
```
cyenyxe: Isn't it a bit extreme to drop **all** the assigned molecules as soon as just one of them has a duplicate?

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/FastaSynonymSequenceReader.java
```diff
@@ -0,0 +1,106 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.ac.ebi.eva.accession.dbsnp.io;
+
+import uk.ac.ebi.eva.accession.core.io.FastaSequenceReader;
+import uk.ac.ebi.eva.accession.dbsnp.contig.ContigMapping;
+import uk.ac.ebi.eva.accession.dbsnp.contig.ContigSynonyms;
+
+public class FastaSynonymSequenceReader {
```
cyenyxe: The purpose of this class is to be a `FastaSequenceReader` with some contig mapping additions, in fact the public API is the same and the implementation of some methods the same. In this particular case I think it would be simpler to override the needed couple of methods to achieve that mapping and inherit the rest of implementations.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/contig/ContigMapping.java
jmmut: Can we rely on the order of the rows on the assembly_reports and guarantee that the first appearance is always the best?

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/FastaSynonymSequenceReader.java
jmmut: Before the refactor I thought delegation would look better than inheritance, but now I agree that inheritance looks good.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/contig/ContigMapping.java
cyenyxe: All the reports we have seen so far list the chromosomes first, but can't be 100% sure if that is always the case.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubmittedVariantRenormalizationProcessor.java
```diff
@@ -97,7 +97,7 @@ private boolean areContextAndLastNucleotideEqual(ISubmittedVariant variant) {
 
     private char getContextBaseInAssembly(ISubmittedVariant variant) {
         long contextPosition = variant.getStart() - 1;
-        String sequence = fastaSequenceReader.getSequence(variant.getContig(), contextPosition, contextPosition);
+        String sequence = fastaSequenceReader.getSequenceToUpperCase(variant.getContig(), contextPosition, contextPosition);
```
cyenyxe: This section shouldn't care about capitalization, so in order to avoid some string transformation I would recommend going back to the plain `getSequence` call.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriter.java
```diff
@@ -159,12 +159,12 @@ private ISubmittedVariant createVariantWithContextBase(ISubmittedVariant normali
             // VCF 4.2 section 1.4.1.4. REF: "the REF and ALT Strings must include the base before the event unless the
             // event occurs at position 1 on the contig in which case it must include the base after the event"
             newStart = normalizedVariant.getStart() + 1;
-            contextBase = fastaSequenceReader.getSequence(normalizedVariant.getContig(), newStart, newStart);
+            contextBase = fastaSequenceReader.getSequenceToUpperCase(normalizedVariant.getContig(), newStart, newStart);
```
cyenyxe: Given that the base is going to be added to the VCF, I think we should reflect what is in the FASTA sequence. VCF accepts any allele capitalization.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/contig/ContigMapping.java
```diff
@@ -20,50 +20,114 @@
...
     public ContigMapping(AssemblyReportReader assemblyReportReader) throws Exception {
-        sequenceNameToSynonyms = new HashMap<>();
-        genBankToSynonyms = new HashMap<>();
-        refSeqToSynonyms = new HashMap<>();
-        ucscToSynonyms = new HashMap<>();
         populateMaps(assemblyReportReader);
     }
 
+    public ContigMapping(List<ContigSynonyms> contigSynonyms) {
+        contigSynonyms.forEach(this::fillContigConventionMaps);
```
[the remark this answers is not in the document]
cyenyxe: I don't find the suggested name particularly clear (what is it convenient for?) and would prefer to keep the current one.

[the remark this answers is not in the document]
pabarcgar: I thought they were "convenience" maps because they were convenient for retrieving the contig synonyms using different kinds of accessions, and that the name was misspelled. But if the meaning is other, then my comment should be ignored.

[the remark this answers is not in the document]
jmmut: they are some kind of convenience maps as they make easier the search. we could have had `List<ContigSynonym>` and make dumb searches, but that would be very slow. Each map allows searching by a different convention (genbank, refseq, seqname, etc).

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/AssemblyCheckerProcessorTest.java
```diff
@@ -186,4 +189,10 @@ public void validReferenceSeqNameInvalidCoordinates() throws Exception {
         assertFalse(processorSeqName.process(input).isAssemblyMatch());
     }
 
+    @Test
+    public void softMaskingShouldNotMatter() throws Exception {
```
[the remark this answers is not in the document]
jmmut: I was using the semantic vocabulary instead of describing sintactic behaviour, which is usually recommended. you can google what soft-masking is, but if we use the words case-insensitive no place in the code will give a clue on why that happens

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/AssemblyReportReader.java
```diff
@@ -78,13 +80,12 @@ public ContigSynonyms read() throws Exception {
...
-            return contigSynonyms;
-        }
-        return null;
+        ContigSynonyms contigSynonyms = new ContigSynonyms(columns[SEQNAME_COLUMN],
+                                                           columns[ASSIGNED_MOLECULE_COLUMN],
+                                                           columns[GENBANK_COLUMN],
+                                                           columns[REFSEQ_COLUMN],
+                                                           columns[UCSC_COLUMN],
+                                                           columns[RELATIONSHIP_COLUMN].equals(IDENTICAL_SEQUENCE));
+        return contigSynonyms;
```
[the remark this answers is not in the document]
jmmut: I sometimes do this to make debugging with breakpoints easier. with lambdas and data flow it's a bit harder to see the values. In this case it's not that important

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/contig/ContigMappingTest.java
```diff
@@ -43,23 +65,75 @@ public void setUp() throws Exception {
         contigMapping = new ContigMapping(fileString);
```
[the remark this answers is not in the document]
jmmut: After the change in policies, the second assembly report is not needed, so now all the tests use the first contigMapping

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/contig/ContigMapping.java
```diff
@@ -57,68 +49,45 @@ public ContigMapping(AssemblyReportReader assemblyReportReader) throws Exception
...
-                assignedMoleculeDuplicates.add(contigSynonyms.getAssignedMolecule());
-            }
-        }
     }
 
     private void normalizeNames(ContigSynonyms contigSynonyms) {
-        contigSynonyms.setSequenceName(removePrefix(contigSynonyms.getSequenceName()));
-        if (contigSynonyms.getAssignedMolecule().equals(NOT_AVAILABLE)) {
+        if (contigSynonyms.getAssignedMolecule().equals(NOT_AVAILABLE)
+                || !ASSEMBLED_MOLECULE.equals(contigSynonyms.getSequenceRole())) {
```
cyenyxe: `NOT_AVAILABLE.equals(contigSynonyms.getAssignedMolecule())` for consistent ordering in both conditions (and avoidance of NPE).

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/contig/ContigMapping.java
```diff
@@ -129,36 +98,24 @@ private void normalizeNames(ContigSynonyms contigSynonyms) {
...
-    private String removePrefix(String contig) {
-        Matcher matcher = PATTERN.matcher(contig);
-        String contigNoPrefix = contig;
-        if (matcher.matches()) {
-            contigNoPrefix = matcher.group(2);
         }
-        return contigNoPrefix;
     }
 
     public ContigSynonyms getContigSynonyms(String contig) {
```
cyenyxe: I assume RefSeq is prioritized because we will most often find those among list of contigs while importing?

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/configuration/ImportDbsnpVariantsStepConfigurationTest.java
```diff
@@ -129,7 +129,8 @@ private void checkRenormalizedVariant(List<DbsnpSubmittedVariantEntity> submitte
                                           List<DbsnpClusteredVariantEntity> clusteredVariants) {
         DbsnpSubmittedVariantEntity submittedVariant = submittedVariants.stream().filter(
                 ss -> ss.getAccession().equals(9999999L)).findFirst().get();
-        SubmittedVariant expectedVariant = new SubmittedVariant("GCF_000002315.4", 9031, "HANDLE_BATCH", "22", 2, "GC",
+        SubmittedVariant expectedVariant = new SubmittedVariant("GCF_000002315.4", 9031, "HANDLE_BATCH", "chr22", 2,
+                                                                "GC",
                                                                 "", 6666666L, false, true, true, false);
```
cyenyxe: All these parameters can fit in a single line.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/contig/ContigMappingTest.java
```diff
@@ -80,43 +74,37 @@ public void removePrefixOnlyAtTheBeginning() {
...
     }
 
     @Test
-    public void getAssignedMoleculeReturnsNullIfThereAreDuplicates() {
-        assertNull(contigMapping.getContigSynonyms(SEQNAME_CONTIG).getAssignedMolecule());
+    public void getSynonymsOfDuplicatedAssignedMolecule() {
+        ContigSynonyms contigSynonyms = contigMapping.getContigSynonyms(ASSIGNED_MOLECULE_CONTIG);
+        assertNotNull(contigSynonyms);
+        assertEquals(ASSEMBLED_MOLECULE_ROLE, contigSynonyms.getSequenceRole());
+        assertNotNull(contigSynonyms.getAssignedMolecule());
```
cyenyxe: Check the rest of synonyms.

jmmut: each of the rest of synonyms fetched from an "assigned molecule" have their own test below. this test is for testing the sequence role.

PR #61, opened 2018-09-04
EVA-1287 Pilot for import into accessioning service - corner cases (unknown orientation again)

jmmut:
we forgot one place where the unknown orientation was producing unexpected behaviour. This PR makes the code:

- assume FORWARD **always** that there's an unknown orientation.
- flag a subsnp as alleles mismatch if there was at least one unknown orientation (or if the reference allele was not found in the list of alleles, which was the only reason to use this flag before).

### DIFF
--- a/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/model/SubSnpNoHgvs.java
+++ b/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/model/SubSnpNoHgvs.java
@@ -88,21 +88,23 @@ public class SubSnpNoHgvs {
         this.taxonomyId = taxonomyId;
         this.anyOrientationUnknown = (subsnpOrientation.getValue() * snpOrientation.getValue()
                 * contigOrientation.getValue()) == 0;
-        Orientation allelesOrientation = Orientation.getOrientation(
-                composeOrientationAssumingForwardIfUnknown(subsnpOrientation, snpOrientation, contigOrientation));
+        Orientation allelesOrientation = composeOrientation(subsnpOrientation, snpOrientation, contigOrientation);
         this.dbSnpVariantAlleles = new DbsnpVariantAlleles(reference, alleles, contigOrientation, allelesOrientation,
                                                            this.dbsnpVariantType);
     }
 
-    private int composeOrientationAssumingForwardIfUnknown(Orientation subsnpOrientation, Orientation snpOrientation,
-                                                           Orientation contigOrientation) {
+    /**
+     * Composes the 3 orientations. If any of those is UNKNOWN orientation, take that particular one as if it were
+     * FORWARD, and then compose it with the others.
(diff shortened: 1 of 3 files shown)


eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/model/SubSnpNoHgvs.java
```diff
@@ -86,12 +86,24 @@ public SubSnpNoHgvs(Long ssId, Long rsId, String reference, String alleles, Stri
...
+        this.anyOrientationUnknown = (subsnpOrientation.getValue() * snpOrientation.getValue()
+                * contigOrientation.getValue()) == 0;
         Orientation allelesOrientation = Orientation.getOrientation(
-                subsnpOrientation.getValue() * snpOrientation.getValue() * contigOrientation.getValue());
+                composeOrientationAssumingForwardIfUnknown(subsnpOrientation, snpOrientation, contigOrientation));
         this.dbSnpVariantAlleles = new DbsnpVariantAlleles(reference, alleles, contigOrientation, allelesOrientation,
                                                            this.dbsnpVariantType);
     }
 
+    private int composeOrientationAssumingForwardIfUnknown(Orientation subsnpOrientation, Orientation snpOrientation,
```
cyenyxe: I think the name is somewhat misleading because it could mean that as soon as one of the flags is unknown we return `Orientation.UNKNOWN`. It think it would be better named as `composeOrientation` and clarify it via method documentation.

Also, is there any benefit on it not returning the result from the `Orientation.getOrientation()` method?

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessorTest.java
```diff
@@ -301,6 +302,23 @@ public void transformSnpReverseContigRsAndSs() throws Exception {
...
+                                                     CONTIG_START, DbsnpVariantType.DIV, Orientation.FORWARD,
+                                                     Orientation.REVERSE, Orientation.UNKNOWN, false, false, false,
+                                                     false, CREATED_DATE, CREATED_DATE, TAXONOMY);
+
+        List<DbsnpSubmittedVariantEntity> variants = processor.process(subSnpNoHgvs).getSubmittedVariants();
+
+        assertEquals(1, variants.size());
+        assertEquals(reference, variants.get(0).getReferenceAllele());
+        assertEquals("", variants.get(0).getAlternateAllele());
+        assertFalse(variants.get(0).isAllelesMatch());
```
cyenyxe: We could also check that it doesn't affect other flags, or check by example, manually creating the appropriate `DbsnpSubmittedVariantEntity` to compare against.

PR #63, opened 2018-09-07
EVA-1330 Return date when calling the web services

andresfsilva:


eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/ClusteredVariant.java
```diff
@@ -44,6 +44,7 @@
     public ClusteredVariant(IClusteredVariant variant) {
         this(variant.getAssemblyAccession(), variant.getTaxonomyAccession(), variant.getContig(), variant.getStart(),
              variant.getType(), variant.isValidated());
+        this.createdDate = variant.getCreatedDate();
```
cyenyxe: I think it could be a more intuitive interface if we added the created date to the list of parameter in the other constructor. The default will be `null` many times, but all the constructors will behave the same way, instead of one setting the date and the other don't.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/processors/VariantProcessor.java
```diff
@@ -59,6 +59,6 @@ public ISubmittedVariant process(final IVariant variant) throws Exception {
                                     SUPPORTED_BY_EVIDENCE,
                                     MATCHES_ASSEMBLY,
                                     ALLELES_MATCH,
-                                    VALIDATED);
+                                    VALIDATED, null);
```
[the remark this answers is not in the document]
andresfsilva: `IVariant` does not have a createdDate field

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToClusteredVariantProcessor.java
```diff
@@ -71,7 +71,7 @@ public DbsnpClusteredVariantEntity process(SubSnpNoHgvs subSnpNoHgvs) throws Exc
                                                         variantRegion.getChromosome(),
                                                         variantRegion.getStart(),
                                                         variantType,
-                                                        subSnpNoHgvs.isSnpValidated());
+                                                        subSnpNoHgvs.isSnpValidated(), null);
```
cyenyxe: This should obtain the creation date from the `subSnpNoHgvs` object. The logic should be the opposite of that in [this method](https://github.com/EBIvariation/eva-accession/pull/63/files#diff-41d0f41cbc7c1968137ceba50d080afaR119), because we would want to retrieve the RS creation time first, and use the SS as a fallback only if the RS doesn't have one.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/SubSnpNoHgvsToDbsnpVariantsWrapperProcessor.java
```diff
@@ -85,7 +85,7 @@ private void addSubmittedVariantEntity(SubSnpNoHgvs subSnpNoHgvs,
         DbsnpSubmittedVariantEntity submittedVariantEntity = new DbsnpSubmittedVariantEntity(subSnpNoHgvs.getSsId(),
                                                                                              hash, submittedVariant,
                                                                                              1);
-        submittedVariantEntity.setCreatedDate(getCreatedDate(subSnpNoHgvs));
+        submittedVariantEntity.setCreatedDate(submittedVariant.getCreatedDate());
```
cyenyxe: This is the same thing as in the other hierarchy. You can add a new argument to the constructor, ideally before `version` to preserve ordering to an extent, then get it from the submitted variant **inside the constructor** instead of via an additional setter call.

This will make sure that any constructed entity is fully correct, instead of depending on an external user to know that the creation date needs to be set afterwards.

eva-accession-ws/src/test/java/uk/ac/ebi/eva/accession/ws/SubmittedVariantsRestControllerTest.java
```diff
@@ -130,4 +121,25 @@ private void assertDefaultFlags(
...
+    public void testGetVariantsController() {
+        List<Long> identifiers = generatedAccessions.stream()
+                                                    .map(acc -> acc.getAccession())
+                                                    .collect(Collectors.toList());
+
+        List<AccessionResponseDTO<SubmittedVariant, ISubmittedVariant, String, Long>> getVariantsResponse =
+                controller.get(identifiers);
+
+        assertEquals(2, getVariantsResponse.size());
+        assertDefaultFlags(getVariantsResponse);
```
cyenyxe: Shouldn't the non-null created date be checked here too?

PR #68, opened 2018-09-25
EVA-1370 Unified variant summary functions

cyenyxe:
dbSNP and EVA variants now share a single summary function, based on the reference sequence and not reference sequence + taxonomy.

This means there will be no ambiguity depending on whether the taxonomy is considered, which makes things easier for us in the future when clustering across dbSNP and EVA submissions. On the other hand, it will cause some goat/sheep IDs to be merged into others.


eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/summary/ClusteredVariantSummaryFunction.java
```diff
@@ -21,7 +21,7 @@
 
 import java.util.function.Function;
 
-public class DbsnpClusteredVariantSummaryFunction implements Function<IClusteredVariant, String> {
+public class ClusteredVariantSummaryFunction implements Function<IClusteredVariant, String> {
```
pabarcgar: Taxonomy is still part of the RS summary function

PR #69, opened 2018-09-26
EVA-1371 Insufficient consistency in mongodb

jmmut:
Decided to use write concern "majority" (in code) and read preference "primary preferred" (in parameters)

### DIFF
--- a/eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/configuration/MongoConfiguration.java
+++ b/eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/configuration/MongoConfiguration.java
@@ -17,6 +17,7 @@ package uk.ac.ebi.eva.accession.core.configuration;
 
 import com.mongodb.MongoClient;
 import com.mongodb.MongoClientOptions;
+import com.mongodb.ReadConcern;
 import com.mongodb.ReadPreference;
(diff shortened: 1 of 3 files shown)


eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/configuration/MongoConfiguration.java
```diff
@@ -56,7 +58,9 @@ public MongoClient mongoClient(MongoProperties properties, ObjectProvider<MongoC
         } else {
             mongoClientOptionsBuilder = new MongoClientOptions.Builder();
         }
-        mongoClientOptions = mongoClientOptionsBuilder.readPreference(ReadPreference.valueOf(readPreference)).build();
+        mongoClientOptions = mongoClientOptionsBuilder.readPreference(ReadPreference.valueOf(readPreference))
```
cyenyxe: Shall we just hardcode the read preference to `primaryPreferred`? The error message recommends using those settings anyway so...

jmmut: this mongo configuration is reused in the eva-accession-import, the eva-accession-pipeline and the eva-accession-ws. in the pipeline it might make sense as well, but I wouldn't say so for the webservices. I would keep it as it is

PR #73, opened 2018-10-04
add eva-accession-import/README.md

jmmut:


eva-accession-import/README.md
```diff
@@ -0,0 +1,25 @@
+# EVA accession import
+
+This module can be used to import accessions (SS and RS) from a dbsnp mirror into the accessioning service at EVA. In https://github.com/EBIvariation/eva-accession/wiki there's some documentation about the transformations done.
```
cyenyxe: `dbSNP mirror`

`Documentation about the transformations applied can be found in https://github.com/EBIvariation/eva-accession/wiki`

eva-accession-import/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+The program can be run with a plain `java -jar eva-accession-import-0.2.0-SNAPSHOT.jar` from a directory with an `application.properties` such as the one under `src/main/resources`.
+
+## Parameters
+
+Tips to fill some parameters:
+
+### `parameters.assemblyAccession`
+
+Look in the dbsnp mirror, in the contiginfo table, the columns "asm_acc" and "asm_version".
```
cyenyxe: `From the dbSNP mirror, concatenate columns 'asm_acc' and 'asm_version' in the 'contiginfo' table.`

eva-accession-import/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+### `parameters.assemblyName`
+
+Look in the dbsnp mirror, in the contiginfo table, the column "group_label".
```
cyenyxe: `From the dbSNP mirror, use column 'group_label' in the 'contiginfo' table.`

eva-accession-import/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+### `parameters.assemblyReportUrl`
+
+Use URLs such as `ftp://ftp.ncbi.nlm.nih.gov/genomes/all/GCF/000/001/735/GCF_000001735.3_TAIR10/GCF_000001735.3_TAIR10_assembly_report.txt`, using the assembly accession (often the GCF one) that you used in `parameters.assemblyAccession`. You can use the equivalent GCA version in most cases, but check if there's any difference between them.
```
cyenyxe:
```
Use URLs such as `ftp://ftp.ncbi.nlm.nih.gov/genomes/all/GCF/000/001/735/GCF_000001735.3_TAIR10/GCF_000001735.3_TAIR10_assembly_report.txt`, pointing to an assembly report.

The recommended assembly to use is the GCF one, and its accession must match the one used in `parameters.assemblyAccession`. You can use the equivalent GCA version in most cases, but please first check there are no differences between them, such an MT contig only present in the GCF assembly.
```

eva-accession-import/README.md
```diff
@@ -0,0 +1,25 @@
+# EVA accession import
+
+This module can be used to import accessions (SS and RS) from a dbsnp mirror into the accessioning service at EVA. In https://github.com/EBIvariation/eva-accession/wiki there's some documentation about the transformations done.
+
+The program can be run with a plain `java -jar eva-accession-import-0.2.0-SNAPSHOT.jar` from a directory with an `application.properties` such as the one under `src/main/resources`.
```
cyenyxe: > java -jar eva-accession-import-\<version\>.jar

eva-accession-import/README.md
jmmut: reworded your second paragraph into `parameters.assemblyAccession`. What I wanted to say here is that the assembly reports for GCF and GCA should be identical. we haven't seen any difference ever.

andresfsilva:
We only have README for the module `eva-accession-import-automation` as everything else is in confluence. The documentation included in this PR as it is right now is outdated so if we want to include this README for this module it will need to be updated and check if we will do the same for other modules.

PR #80, opened 2018-10-25
EVA-1387 Exclude variants with explicit alleles invalid in VCF

andresfsilva:


eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/steps/processors/ExcludeInvalidVariantsProcessor.java
```diff
@@ -0,0 +1,37 @@
...
+import uk.ac.ebi.eva.commons.core.models.IVariant;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Filters variants with invalid alleles.
+ *
+ * VCF format only accepts reference and alternate alleles formed by A, C, G, R or N letters in upper or lower case.
+ * If one of the alleles has a different character, this processor will return NULL so tha variant can be ignored.
```
jmmut:
```suggestion
 * If one of the alleles has a different character, this processor will return null so that variant can be ignored.
```

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/steps/processors/ExcludeInvalidVariantsProcessor.java
```diff
@@ -0,0 +1,37 @@
...
+    public IVariant process(IVariant variant) throws Exception {
+        if (variant.getReference().isEmpty() || variant.getAlternate().isEmpty()) {
+            throw new IllegalArgumentException(REFERENCE_AND_ALTERNATE_ALLELES_CANNOT_BE_EMPTY);
+        }
+
+        Matcher matcher = ALLELES_PATTERN.matcher(variant.getReference() + variant.getAlternate());
+        if(matcher.matches()) {
+            return variant;
+        }
+        return null;
```
jmmut: what do you think about logging how many variants were filtered out? Maybe you can inherit ItemStreamProcessor to have a `close` method to log there the counts

jmmut: ok, there's no ItemStreamProcessor, we'll need a a listener or something else

jmmut: something like 
```
public class SkipLoggerListener<I, O> extends StepListenerSupport<I, O> {

    private static final Logger logger = LoggerFactory.getLogger(SkipLoggerListener.class);

    @Override
        logger.info("Processors filtered out {} variants", stepExecution.getProcessSkipCount());
        return stepExecution.getExitStatus();
    }
}
```

eva-accession-release/src/test/java/uk/ac/ebi/eva/accession/release/steps/processors/ExcludeInvalidVariantsProcessorTest.java
```diff
@@ -0,0 +1,148 @@
...
+        thrown.expectMessage(REFERENCE_AND_ALTERNATE_ALLELES_CANNOT_BE_EMPTY);
+        assertNull(processor.process(variant));
+    }
+
+    @Test
+    public void alternateAlleleEmpty() throws Exception {
+        IVariant variant = newVariant(VALID_ALLELE, EMPTY_ALLELE);
+        thrown.expect(IllegalArgumentException.class);
+        thrown.expectMessage(REFERENCE_AND_ALTERNATE_ALLELES_CANNOT_BE_EMPTY);
+        assertNull(processor.process(variant));
```
jmmut: a small detail: if `process` is going to throw an exception, the `assertNull` is misleading and unused.

here and in the previous test

PR #81, opened 2018-11-01
EVA-1384 Import cow_9913 into accessioning service (allow several merges)

jmmut:
## Before

Previously the code assumed that a single clustered variant could not be merged into several other clustered variants. However this design had 2 problems:

- Some species have lots of equivalent variants (such as cow ) and the code will have to deal with this situation more times that we expected initially (we expected 0 times).
- A single clustered variant could actually be merged into several others (and would go unnoticed) if the clustered variant had 2 or more entries that are not received near each other (not in the same chunk). This can not be avoided unless querying the operations collection is queried for each clustered variant to process.

This was detected with:
```
java.lang.IllegalStateException: Clustered variant rs136611820 was merged into several other clustered variants with the same hash: [42568024, 42568025]
```
and the corresponding tests were added that replicate the situation with rs136611820.

## Now

This pull request removes the constraint of having only one merge for each clustered variant. There are some policies to handle the consequences of that:

- When a clustered variant is merged into another, the submitted variants associated have their `rs` field updated to the "merged into" clustered variant. If there are multiple "merged into" variants, the one that is active is used. In other words, the one that is in the main collection  dbsnpClusteredVariantEntity is used, as there will be only one with the hash that made collision with the clustered variant being handled.
  - A clustered variant being merged into several other accessions can only happen in 3 scenarios:
    - Same clustered variant accession, different hashes, hash collisions with other declustered variants. There will be only one document written in the declustered collection **per hash**, so one merge operation is issued per hash, and the clustered variants can be identified in the history collection with the hash. 
    - Same accession different hashes, hash collisions with other active variants. Similar to the previous one, issue one merge operation per hash, and use the clustered variant hash to identify the wrapper where the submitted variants should get the `rs` field updated.
    - Same accession, same hash, one collision in the active collection and one collision in the declustered collection. In this case the active variant with that hash will be queried, and the active submitted variants will get that active accession, and the declustered variants will stay declustered, i.e. `rs=null`. It can't happen that the same wrapper (same hash and same accession) is merged into several other accessions, unless there are mixed declustered and non-declustered, because a hash is unique within a collection; so even dealing with repeated wrappers, a "distinct" will be performed on the operations to issue each distinct merge operation only once.
- Given a row (one clustered variant and a list of submitted variants), only update the `rs` field of submitted variants that are **not** declustered (and write for each non-declustered submitted variant the operation updating the `rs` field). In other words, don't replace `rs:null` with `rs:<mergedInto>`, and for the declustered submitted variants, write only declustering operations.
- In the future, when we query the history of an RS, we might receive **several** operations for merging that RS into others. This result will be expected, and to disambiguate the correct "mergedInto" RS, it has to be queried in the main collection, querying by hash, and it should match the `rs` field of the submitted variants of the original RS. Note that an RS might have different hashes (for instance, several locations, or several types), so several different hashes might be active at the same time, that's another reason to query by hash.

### DIFF
--- a/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
+++ b/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
@@ -271,7 +271,7 @@ public class DbsnpVariantsWriter implements ItemWriter<DbsnpVariantsWrapper> {
         for (DbsnpSubmittedVariantEntity submittedVariant : wrapper.getSubmittedVariants()) {
             Long mergedInto = replacements.get(Pair.of(wrapper.getClusteredVariant().getHashedMessage(),
                                                        wrapper.getClusteredVariant().getAccession()));
-            if (mergedInto != null) {
+            if (mergedInto != null && submittedVariant.getClusteredVariantAccession() != null) {
                 operations.add(buildOperation(submittedVariant, mergedInto));
                 submittedVariant.setClusteredVariantAccession(mergedInto);
             }
--- a/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
+++ b/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
@@ -286,15 +286,22 @@ public class DbsnpVariantsWriter implements ItemWriter<DbsnpVariantsWrapper> {
                                                      DbsnpClusteredVariantEntity activeClusteredVariant) {
         String activeVariantMessage;
         if (activeClusteredVariant == null) {
-            activeVariantMessage = ". There is no active variant with that hash.";
+            activeVariantMessage = "There is no active variant with that hash.";
         } else {
-            activeVariantMessage = ". The active variant with that hash is rs" + activeClusteredVariant.getAccession();
+            activeVariantMessage =
+                    "However, there is an active variant with that hash and accession rs"
+                            + activeClusteredVariant.getAccession() + ".";
+
         }
+        String accessionsMergedInto = mergedIntoList.stream()
+                                                    .map(accession -> "rs" + accession)
+                                                    .collect(Collectors.joining(", "));
         throw new IllegalStateException(
                 "Clustered variant rs" + hashAndAccession.getSecond() + " was merged into several other "
-                        + "clustered variants with the same hash: " + mergedIntoList + " and none of them"
-                        + " is active, i.e. present in the collection for "
-                        + DbsnpClusteredVariantEntity.class.getSimpleName() + activeVariantMessage);
+                        + "clustered variants (" + accessionsMergedInto + ") because all have the same hash ("
+                        + hashAndAccession.getFirst() + ") but none of those clustered variants is active, i.e. "
+                        + "present in the collection for " + DbsnpClusteredVariantEntity.class.getSimpleName()
+                        + ". " + activeVariantMessage);
     }
 
     private void updateClusteredVariantAccessionsInSubmittedVariants(DbsnpVariantsWrapper wrapper,
--- a/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
+++ b/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
@@ -412,7 +412,7 @@ public class DbsnpVariantsWriter implements ItemWriter<DbsnpVariantsWrapper> {
                     .stream()
                     .filter(v -> v.getHashedMessage().equals(hash)
                             && !v.getAccession().equals(mergedInto.getAccession())
-                            && !isAlreadyMerged(v.getAccession()))
+                            && !isAlreadyMergedInto(v, mergedInto))
                     .map(origin -> mergeOperationFactory.apply(origin, mergedInto))
                     .collect(Collectors.toList());
         }
@@ -439,9 +439,13 @@ public class DbsnpVariantsWriter implements ItemWriter<DbsnpVariantsWrapper> {
                                       .values();
         }
 
-        private boolean isAlreadyMerged(Long accession) {
-            List<OPERATION_ENTITY> merges = operationRepository.findAllByAccession(accession);
-            return merges.stream().anyMatch(operation -> operation.getEventType().equals(EventType.MERGED));
+        private boolean isAlreadyMergedInto(ENTITY original, ENTITY mergedInto) {
+            List<OPERATION_ENTITY> merges = operationRepository.findAllByAccession(original.getAccession());
+            return merges.stream().anyMatch(
+                    operation ->
+                            operation.getEventType().equals(EventType.MERGED)
+                                    && mergedInto.getAccession().equals(operation.getMergedInto())
+                                    && original.getHashedMessage().equals(operation.getInactiveObjects().get(0).getHashedMessage()));
         }
     }
 }
--- a/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
+++ b/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
@@ -288,6 +288,9 @@ public class DbsnpVariantsWriter implements ItemWriter<DbsnpVariantsWrapper> {
 
         Long accession = originalSubmittedVariant.getAccession();
         DbsnpSubmittedVariantOperationEntity operation = new DbsnpSubmittedVariantOperationEntity();
+
+        // Note the next null in accessionIdDestiny. We are not merging the submitted variant into
+        // clusteredVariantMergedInto. We are updating the submitted variant, changing its rs field
         operation.fill(EventType.UPDATED, accession, null, reason, Collections.singletonList(inactiveEntity));
         return operation;
     }
--- a/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
+++ b/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
@@ -312,8 +312,8 @@ public class DbsnpVariantsWriter implements ItemWriter<DbsnpVariantsWrapper> {
 
     private DbsnpSubmittedVariantOperationEntity buildOperation(DbsnpSubmittedVariantEntity originalSubmittedVariant,
                                                                 Long clusteredVariantMergedInto) {
-        DbsnpSubmittedVariantInactiveEntity inactiveEntity =
-                new DbsnpSubmittedVariantInactiveEntity(originalSubmittedVariant);
+        DbsnpSubmittedVariantInactiveEntity inactiveEntity = new DbsnpSubmittedVariantInactiveEntity(
+                originalSubmittedVariant);
 
         Long originalClusteredVariant = originalSubmittedVariant.getClusteredVariantAccession();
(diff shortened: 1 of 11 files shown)


eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
```diff
@@ -255,23 +262,48 @@ private void writeSubmittedVariantsAndOperations(List<? extends DbsnpVariantsWra
...
             }
-
-            originalToNewAccessions.put(hashAndAccession, mergedIntoList.get(0));
         });
         return originalToNewAccessions;
     }
 
+    private void throwSeveralInactiveMergesException(Pair<String, Long> hashAndAccession, List<Long> mergedIntoList,
+                                                     DbsnpClusteredVariantEntity activeClusteredVariant) {
+        String active;
```
sundarvenkata-EBI: Consider renaming to "activeVariantMessage"

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
```diff
@@ -255,23 +262,48 @@ private void writeSubmittedVariantsAndOperations(List<? extends DbsnpVariantsWra
...
+                        hashAndAccession.getFirst(), DbsnpClusteredVariantEntity.class);
+
+                if (activeClusteredVariant == null || !mergedIntoList.contains(activeClusteredVariant.getAccession())) {
+                    throwSeveralInactiveMergesException(hashAndAccession, mergedIntoList, activeClusteredVariant);
+                } else {
+                    originalToNewAccessions.put(hashAndAccession, activeClusteredVariant.getAccession());
+                    logger.debug("Clustered variant rs{} was merged into several other clustered variants with the "
+                                         + "same hash: {}. The accession rs{} was chosen as "
+                                         + "replacement because it's active, i.e. present in the collection for {}",
+                            hashAndAccession.getSecond(), mergedIntoList, activeClusteredVariant.getAccession(),
```
sundarvenkata-EBI: Will this code path be exercised for rs136611820 in EVA-1384? I did `grep "chosen as replacement" cow_9913_load_to_accessioning_20181101120650.log` and I couldn't find anything.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
```diff
@@ -255,23 +262,48 @@ private void writeSubmittedVariantsAndOperations(List<? extends DbsnpVariantsWra
                     .collect(Collectors.toList());
 
             if (mergedIntoList.size() > 1) {
-                throw new IllegalStateException(
-                        "Clustered variant rs" + hashAndAccession.getSecond() + " was merged into several other "
-                                + "clustered variants with the same hash: " + mergedIntoList);
+                DbsnpClusteredVariantEntity activeClusteredVariant = mongoTemplate.findById(
+                        hashAndAccession.getFirst(), DbsnpClusteredVariantEntity.class);
+
+                if (activeClusteredVariant == null || !mergedIntoList.contains(activeClusteredVariant.getAccession())) {
```
sundarvenkata-EBI: Is it a plausible scenario to get an active clustered variant with the hash appear after the current batch?

jmmut: yes, but in any case I don't see any problem.
- if the clustered variant that appears later is the 3rd one with the same hash, it will make collision and be handled in the same way as the 2nd.
- if the clustered variant that appears later is the 2nd one, and now we are handling the 1st one, this 1st one won't make collision and will become the active clustered variant for its hash.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
jmmut: yes, but only if the debug level is debug. the default is info, I think.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriterTest.java
```diff
@@ -437,53 +485,58 @@ public void mergeDuplicateSubmittedVariants() throws Exception {
...
-                    assertEquals(submittedVariantEntity.isAssemblyMatch(), inactiveEntity.isAssemblyMatch());
-                    assertEquals(submittedVariantEntity.isAllelesMatch(), inactiveEntity.isAllelesMatch());
-                    assertEquals(submittedVariantEntity.isValidated(), inactiveEntity.isValidated());
-                });
+                    assertEquals(mergedInto.isSupportedByEvidence(),
+                                 inactiveEntity.isSupportedByEvidence());
+                    assertEquals(mergedInto.isAssemblyMatch(), inactiveEntity.isAssemblyMatch());
+                    assertEquals(mergedInto.isAllelesMatch(), inactiveEntity.isAllelesMatch());
+                    assertEquals(mergedInto.isValidated(), inactiveEntity.isValidated());
+                    return 1;
```
sundarvenkata-EBI: I would like to understand the rationale for returning a value. Is it to count the number of entities that made it past all the asserts? Do assert failures inside the map break the control flow?

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriterTest.java
```diff
@@ -914,6 +954,374 @@ public void declusterRs638662487() throws Exception {
...
+        // could be 3 operations total? 2 copies for clusteredVariantEntity3
+
+        assertEquals(0, mongoTemplate.count(new Query(), DBSNP_CLUSTERED_VARIANT_DECLUSTERED_COLLECTION_NAME));
+    }
+
+    /**
+     * Clustered variant rs136611820 was merged into several other clustered variants with the same hash: [42568024,
+     * 42568025]
+     * <p>
+     * This happened because rs136611820 had several locations. One matched the location of rs42568024 and other matched
```
sundarvenkata-EBI: "had several locations" is misleading as we discussed offline. Re-phrase to "had locations that matched with that of multiple ss IDs".

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriterTest.java
```diff
@@ -914,6 +954,374 @@ public void declusterRs638662487() throws Exception {
...
+     */
+    @Test
+    public void rs136611820() throws Exception {
+        // given
+        DbsnpVariantsWrapper wrapper1 = new DbsnpVariantsWrapper();
+        DbsnpVariantsWrapper wrapper2 = new DbsnpVariantsWrapper();
+        DbsnpVariantsWrapper wrapper3 = new DbsnpVariantsWrapper();
+        DbsnpVariantsWrapper wrapper4 = new DbsnpVariantsWrapper();
+
+        setupRs136611820(wrapper1, wrapper2, wrapper3, wrapper4);
```
sundarvenkata-EBI: Could you please ensure that the contig start positions do not differ for the ss IDs during the setup? While I understand that they are irrelevant, they are an unneeded distraction when debugging.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriterTest.java
```diff
@@ -914,6 +954,374 @@ public void declusterRs638662487() throws Exception {
...
+                                                                               clusteredVariantEntity1.getAccession());
+        assertSubmittedVariantsStored(2, submittedVariantEntity1, expectedSubmittedVariantEntity3);
+        assertSubmittedVariantsHaveActiveClusteredVariantsAccession(wrapper.getClusteredVariant().getAccession(),
+                                                                    expectedSubmittedVariantEntity2,
+                                                                    expectedSubmittedVariantEntity3);
+        assertSubmittedUpdateOperationsHaveClusteredVariantAccession(5, 1, clusteredVariantAccession2);
+        assertSubmittedUpdateOperationsHaveClusteredVariantAccession(5, 2, clusteredVariantAccession3);
+        assertSubmittedVariantMergeOperationStored(5, 1, submittedVariantEntity1);
+        assertSubmittedVariantMergeOperationStored(5, 1, submittedVariantEntity3);
+        // could be 3 operations total? 2 copies for clusteredVariantEntity3
```
andresfsilva: Delete line

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriterTest.java
jmmut: yes, it's for counting how many match the mergeInto field. would it be more clear to use 2 mongoTemplate queries, the second with  `query(where("mergeInto").is(mergedInto.getAccession)` and use a foreach without return?

jmmut: and yes, assert failures inside the map break the control flow. what do you think should be checked first? the count or the fields? I guess the counts first is more correct

sundarvenkata-EBI: I agree that validating the counts first would be better.

eva-accession-import/src/test/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriterTest.java
```diff
@@ -345,19 +317,56 @@ public void repeatedClusteredVariantsPartiallyDeclustered() throws Exception {
         assertClusteredVariantDeclusteredStored(1, wrapper1);
     }
 
+    @Test
+    public void mergedClusteredVariantsPartiallyDeclustered() throws Exception {
+        boolean allelesMatch = false;
+        SubmittedVariant submittedVariant1 = buildSubmittedVariant(CLUSTERED_VARIANT_ACCESSION_1);
+        submittedVariant1.setAllelesMatch(allelesMatch);
```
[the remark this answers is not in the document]
jmmut: refactored a bit so that there's a helper `decluster` that creates the operation and sets the link to null

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpVariantsWriter.java
```diff
@@ -255,23 +262,48 @@ private void writeSubmittedVariantsAndOperations(List<? extends DbsnpVariantsWra
...
+            activeVariantMessage = ". There is no active variant with that hash.";
+        } else {
+            activeVariantMessage = ". The active variant with that hash is rs" + activeClusteredVariant.getAccession();
+        }
+        throw new IllegalStateException(
+                "Clustered variant rs" + hashAndAccession.getSecond() + " was merged into several other "
+                        + "clustered variants with the same hash: " + mergedIntoList + " and none of them"
+                        + " is active, i.e. present in the collection for "
+                        + DbsnpClusteredVariantEntity.class.getSimpleName() + activeVariantMessage);
+    }
```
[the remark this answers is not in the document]
sundarvenkata-EBI: >>but there's an active activeClusteredVariant and this is its RS accession.

Let's say exactly that. 
activeVariantMessage = ". But there is an active variant with that hash: rs" + activeClusteredVariant.getAccession();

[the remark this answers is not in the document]
jmmut: improved a bit this line and the main exception message

PR #82, opened 2018-11-02
EVA-1358 Get list of variants in an assembly

jmmut:
Perform the aggregation (with the `$lookup`) to query clustered variants.

**Note**: In another ticket, we need to do a processor to exclude clustered variants that don't have any submitted variants.

The raw MongoDB java driver was used, as our version of Spring Data MongoDB returned the results in a list, not in a cursor. Newer versions of Spring Data MongoDB have what we want, but we would have to migrate to Spring Boot 2.

PR #83, opened 2018-11-06
EVA-1378 - Exclude clustered variants without submitted variants

sundarvenkata-EBI:


jmmut:
This is no longer necessary after the grouping (chr + pos + ref + alt) made in #82

PR #85, opened 2018-11-13
EVA-1385 Step and job setup

jmmut:


eva-accession-release/src/test/resources/test-data/dbsnpClusteredVariantEntity.json
```diff
@@ -18,7 +18,7 @@
       "asm": "GCF_000409795.2",
       "tax": 60711,
       "contig": "CM001941.2",
-      "start": NumberLong(13247923),
+      "start": NumberLong(13),
       "type": "SNV",
       "accession": NumberLong(869927931),
       "version": 1,
```
sundarvenkata-EBI: Could you please add more kinds of variants (like INDEL, NAMED, STR etc.,)?

eva-accession-release/src/test/java/uk/ac/ebi/eva/accession/release/configuration/CreateReleaseStepConfigurationTest.java
```diff
@@ -0,0 +1,146 @@
...
+    @Test
+    public void basicJobCompletion() throws Exception {
+        JobExecution jobExecution = jobLauncherTestUtils.launchStep(CREATE_RELEASE_STEP);
+        assertEquals(BatchStatus.COMPLETED, jobExecution.getStatus());
+    }
+
+    @Test
+    public void variantsWritten() throws Exception {
+        JobExecution jobExecution = jobLauncherTestUtils.launchStep(CREATE_RELEASE_STEP);
+        assertEquals(BatchStatus.COMPLETED, jobExecution.getStatus());
```
sundarvenkata-EBI: Extract these two statements to share with basicJobCompletion()?

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/runner/AccessionReleaseJobLauncherCommandLineRunner.java
```diff
@@ -0,0 +1,151 @@
+/*
```
sundarvenkata-EBI: Since this program is the same as the commandlinerunner from other modules, could we share code across the classes?

eva-accession-release/src/test/java/uk/ac/ebi/eva/accession/release/configuration/CreateReleaseStepConfigurationTest.java
jmmut: they are not the same. the step test uses `launchStep` and the job one uses `launchJob`. I  extracted the lines and reused 4 times in the test, although I think it's quite clear as it is.

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/runner/AccessionReleaseJobLauncherCommandLineRunner.java
jmmut: the thing is that several pipelines might need different CLI features. these two are the same, but use different parameters, so we could make the class generic on the InputParameters, and if we need to add something like a CLI parameter (not on the properties file), we will have to extend the class and make constructors, overload one method or two, and at that point it will be exactly as it is now.

What I mean is, it's good to extract the same behaviours to a common place if you want to keep the behaviours parallel, but I don't think that's the scenario here.

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/configuration/AccessionReleaseJobConfiguration.java
```diff
@@ -0,0 +1,31 @@
+package uk.ac.ebi.eva.accession.release.configuration;
```
cyenyxe: Copyright header.

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/configuration/BeanNames.java
```diff
@@ -2,7 +2,17 @@
 
 public class BeanNames {
 
-    public static final String EXCLUDE_VARIANTS_LISTENER = "EXCLUDE_VARIANTS_LISTENER";
+    public static final String ACCESSION_RELEASE_JOB = "ACCESSION_RELEASE_JOB";
+
+    public static final String CREATE_RELEASE_STEP = "CREATE_RELEASE_STEP";
```
cyenyxe: RELEASE_MAPPED_ACTIVE_VARIANTS_STEP? This seems more specific considering we will create reports for variants that are:

* Active, merged and deprecated
* Mapped and unmapped

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/listeners/GenericProgressListener.java
```diff
@@ -0,0 +1,95 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.batch.core.ExitStatus;
+import org.springframework.batch.core.StepExecution;
+import org.springframework.batch.core.listener.StepListenerSupport;
+import org.springframework.batch.core.scope.context.ChunkContext;
+
+import java.util.List;
+
+public class GenericProgressListener<I, O> extends StepListenerSupport<I, O> {
```
cyenyxe: Is this a direct copy from [the one](https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/listeners/GenericProgressListener.java) in eva-accession-import? Could it be moved to eva-accession-core then, so it can be reused by multiple pipelines?

jmmut: yes, I got confused here

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/io/FastaSynonymSequenceReader.java
```diff
@@ -30,6 +30,22 @@ public FastaSynonymSequenceReader(ContigMapping contigMapping, Path fastaPath) t
...
+        if (contigSynonyms == null) {
+            throw new IllegalArgumentException(
+                    "Contig '" + contig + "' not found in the assembly report");
+        }
+
+        return super.doesContigExist(contigSynonyms.getSequenceName())
+                || super.doesContigExist(contigSynonyms.getGenBank())
+                || super.doesContigExist(contigSynonyms.getRefSeq())
+                || super.doesContigExist(contigSynonyms.getUcsc())
+                || super.doesContigExist(contigSynonyms.getAssignedMolecule());
```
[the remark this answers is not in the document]
jmmut: The issue you mention won't be a problem because `ContigMapping` will set the assigned molecule to null if the sequence-role is not "assembled-molecule". So only one row (the correct one) will match `1`.

however, I added the condition to check for synonyms only if genbank and refseq are identical, to match the logic in `getSequence`.

There's an issue pending, where a given assigned-molecule appears several times as "assembled-molecule", but I plan to make that change in the eva-accession-import branch 0.2.x and then forward-port those changes into master

PR #86, opened 2018-11-15
EVA-1324 Import primates_61853 into accessioning service

jmmut:
## Note:
This PR has to be merged after #81 . This PR is only from the commit "contigSynonyms: ignore nonunique assigned-molecule ..." onwards.

This PR introduces 2 changes:
- simplify contigReplacerProcessor: just transform the contig to genbank if they are identical
- ContigSynonyms: ignore non-unique assigned-molecule with "assembled-molecule".


eva-accession-import/src/test/resources/input-files/fasta/fastaWithSoftMasking.fa.fai
```diff
@@ -0,0 +1 @@
+NW_006738765.1	160	128	80	81
```
sundarvenkata-EBI: Not sure I understand. So the FAI file was included when it shouldn't have been?

jmmut:
looks good to me. A more elaborate code would aggregate all those contigs missing from the assembly report in a single set, and logged after the step, but as discussed with @sundarvenkata-EBI, it's not trivial as the ItemProcessor interface doesn't have access to the step execution context, so a custom class (or a custom injection through listeners) would be needed to store those contigs (and maybe also how many variants were present in those contigs).

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/ContigReplacerProcessor.java
```diff
@@ -23,70 +25,49 @@
...
-        } else {
-            // genbank is not identical to refseq and the assembly is not genbank, so
-            // we must keep the original refseq, even if the refseq was not found in the assembly report
+    private boolean isContigPresentInAssemblyReport(SubSnpNoHgvs subSnpNoHgvs, ContigSynonyms contigSynonyms) {
+        boolean contigPresentInAssemblyReport = contigSynonyms != null;
+        if (!contigPresentInAssemblyReport) {
+            logger.warn(
+                    "CONTIG NOT FOUND: Contig '" + subSnpNoHgvs.getContigName() +
+                            "' was not found in the assembly report! Is the " + "assembly accession '" +
+                            assemblyAccession + "' correct?");
```
sundarvenkata-EBI: I observed that this scenario is already covered by https://github.com/jmmut/eva-accession/blob/feature/several-assembled-molecules/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/FastaSynonymSequenceReader.java#L39. I feel that this error belongs in FastaSynonymSequenceReader because it is a potentially reusable class and having a guard against Contig not found in the assembly report is better. @jmmut's opinion is that for this particular error, the ContigReplacer is a better option. @cyenyxe @andresfsilva Any thoughts?

jmmut: I'm not sure I understand your point. My point is that the default behaviour (in FastaSynonymSequenceReader) should be throwing an exception if the contig is not found, but for the case of the contig replacer we know this is not an error so we can catch the exception and only log it. Other usages of the FastaSynonymSequenceReader can rely on the exception-throwing behaviour.

Of course what I suggest does not apply if we decide that the general case should expect missing contigs in the assembly report, which is kind of true if we take into account all the strange contigs. However, I don't think this should be the general case because we will accept INSDC sequences only in most parts of EVA.

jmmut:
ok, several things have happened since this PR was created. We decided to import all dbsnp variants as if they were in genbank, both assemblies and contigs, so there's a small part here that is outdated. however this includes changes that are still needed for importing primates and that #96 doesn't include

cyenyxe:
Superseded by #91 and #92.

PR #95, opened 2019-01-30
EVA-1436 Prepare FASTA file with all INSDC sequences in an assembly report

cyenyxe:
Given an assembly report, download the FASTA sequence for all the GenBank accessions listed in it, and build a FASTA file that should cover all those present in the dbSNP database.

This script only works for RefSeq sequences which have GenBank equivalents. It is not possible to download RefSeq-only sequences from ENA, and there is no easy way to retrieve them from NCBI either :disappointed:

### DIFF
--- a/eva-accession-import-automation/create_fasta_from_assembly_report.sh
+++ b/eva-accession-import-automation/create_fasta_from_assembly_report.sh
@@ -20,7 +20,7 @@ do
 
     # Download each GenBank accession in the assembly report from ENA into a separate file
     # Delete the accession prefix from the header line
-    wget -q -O - "https://www.ebi.ac.uk/ena/data/view/${genbank_contig}&display=fasta" | sed 's/ENA|.*|//g' > ${genbank_contig}
+    wget -q -O - "https://www.ebi.ac.uk/ena/browser/api/fasta/${genbank_contig}" | sed 's/ENA|.*|//g' > ${genbank_contig}
 
     # If a file has more than one line, then it is concatenated into the full assembly FASTA file
     # (empty sequences can't be indexed)
--- a/eva-accession-import-automation/create_fasta_from_assembly_report.sh
+++ b/eva-accession-import-automation/create_fasta_from_assembly_report.sh
@@ -2,7 +2,7 @@
 
 # Usage: create_fasta_from_assembly_report.sh </path/to/assembly_report.txt>
 
-if [ "$#" -lt 3 ]
+if [ "$#" -ne 3 ]
 then
     echo "Please provide the assembly accession, the path to the assembly report and the output folder."
     exit 1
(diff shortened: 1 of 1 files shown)


eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+# Usage: create_fasta_from_assembly_report.sh </path/to/assembly_report.txt>
+
+if [ "$#" -lt 3 ]
```
jmmut: Not critical, but I usually check it's not equal to the expected number. An extra word between the first parameters will silently blow everything up. I think 
```suggestion
if [ "$#" -ne 3 ]
```
should work

eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -0,0 +1,42 @@
...
+then
+    echo "Please provide the assembly accession, the path to the assembly report and the output folder."
+    exit 1
+fi
+
+assembly_accession=$1
+assembly_report=$2
+output_folder=$3
+
+for i in `grep -v -e "^#" ${assembly_report} | cut -f5`;
```
jmmut: replace `i` for `genbank_contig`? if you do this, do it after fixing the other comments

eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -0,0 +1,42 @@
...
+do
+    echo ${i}
+
+    # Download each GenBank accession in the assembly report from ENA into a separate file
+    # Delete the accession prefix from the header line
+    wget -q -O - "https://www.ebi.ac.uk/ena/data/view/${i}&display=fasta" | sed 's/ENA|.*|//g' > ${i}
+
+    # If a file has more than one line, then it is concatenated into the full assembly FASTA file
+    # (empty sequences can't be indexed)
+    lines=`wc -l ${i} | cut -f1 -d' '`
```
jmmut: no need to read the whole file to check if there are more than two lines, and no need for `cut` if you already ask only for the lines (`-l`).
```suggestion
    lines=`head -n 2 ${i} | wc -l`
```

eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -0,0 +1,42 @@
...
+    if [ $lines -ne 1 ]
+    then
+        cat ${genbank_contig} >> ${output_folder}/${assembly_accession}.fa
+    else
+        echo FASTA sequence not available for ${genbank_contig}
+    fi
+
+    # Check that an accession is present no more than once in the output FASTA file, otherwise it 
+    # means there are unexpected aliases in the assembly report, which need to be checked
+    acc=`head -n 1 ${genbank_contig} | cut -f1 -d' ' | cut -f1 -d'.' | cut -f2 -d'>'`
```
andresfsilva:
```suggestion
    accession=`head -n 1 ${genbank_contig} | cut -f1 -d' ' | cut -f1 -d'.' | cut -f2 -d'>'`
```

eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -0,0 +1,42 @@
...
+    matches=`grep -c "${acc}" ${output_folder}/${assembly_accession}.fa`
```
andresfsilva:
```suggestion
    matches=`grep -c "${accession}" ${output_folder}/${assembly_accession}.fa`
```

PR #100, opened 2019-02-07
EVA-1453 Generate VCF report for release of merged RS IDs

jmmut:
added a step to the single job in eva-accession-release


eva-accession-release/src/test/java/uk/ac/ebi/eva/accession/release/io/MergedVariantMongoReaderTest.java
```diff
@@ -0,0 +1,219 @@
...
+    public void includeAllelesMatchFlag() throws Exception {
+        assertFlagEqualsInAllVariants(ALLELES_MATCH_KEY, true);
+    }
+
+    @Test
+    public void includeEvidenceFlag() throws Exception {
+        assertFlagEqualsInAllVariants(SUPPORTED_BY_EVIDENCE_KEY, true);
+    }
+
+}
```
jmmut: I could add more tests, but I'm not sure we need to put them because it's already tested in the parent class, here we test that the "mergedInto" attribute was picked up correctly, and also the alleles and flags from the joined collection

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/MergedVariantContextWriter.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public MergedVariantContextWriter(File output, String referenceAssembly) {
+        super(output, referenceAssembly);
+    }
+
+    @Override
+    protected Set<VCFHeaderLine> buildHeaderLines() {
+        Set<VCFHeaderLine> vcfHeaderLines = super.buildHeaderLines();
+        vcfHeaderLines.add(new VCFInfoHeaderLine(MERGED_INTO_KEY, 1, VCFHeaderLineType.String,
+                                                 "Active RS id to represent a given variant"));
```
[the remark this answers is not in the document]
cyenyxe: What about

> RS ID that currently represents the variant

?

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/MergedVariantMongoReader.java
```diff
@@ -0,0 +1,239 @@
...
+
+    private static final String ALLELES_MATCH_FIELD = "allelesMatch";
+
+    private static final String SUPPORTED_BY_EVIDENCE_FIELD = "evidence";
+
+    public static final String VARIANT_CLASS_KEY = "VC";
+
+    public static final String STUDY_ID_KEY = "SID";
+
+    public static final String MERGED_INTO_KEY = "A";   // Active TODO: any suggestions on a better name?
```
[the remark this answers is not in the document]
cyenyxe: If we go the "current" way, it could be `CUR` or `CURR`.

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/configuration/CreateMergedReleaseStepConfiguration.java
```diff
@@ -0,0 +1,77 @@
...
+    @Qualifier(PROGRESS_LISTENER)
+    private StepExecutionListener progressListener;
+
+    @Autowired
+    @Qualifier(EXCLUDE_VARIANTS_LISTENER)
+    private StepExecutionListener excludeVariantsListener;
+
+    @Bean(RELEASE_MAPPED_MERGED_VARIANTS_STEP)
+    public Step createMergedReleaseStep(StepBuilderFactory stepBuilderFactory,
+                                          SimpleCompletionPolicy chunkSizeCompletionPolicy) {
```
cyenyxe: Fix indentation.

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/AccessionedVariantMongoReader.java
```diff
@@ -123,8 +122,12 @@ public AccessionedVariantMongoReader(String assemblyAccession, MongoClient mongo
 
     @Override
     public void open(ExecutionContext executionContext) throws ItemStreamException {
+        executeAggregation(DBSNP_CLUSTERED_VARIANT_ENTITY);
```
cyenyxe: Or maybe just `aggregate`?

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/AccessionedVariantMongoReader.java
```diff
@@ -60,35 +59,35 @@
...
 
-    private static final String ACCESSION_FIELD = "accession";
+    protected static final String ACCESSION_FIELD = "accession";
 
     private static final String REFERENCE_ASSEMBLY_FIELD = "asm";
 
-    private static final String STUDY_FIELD = "study";
+    protected static final String STUDY_FIELD = "study";
 
     private static final String CONTIG_FIELD = "contig";
```
cyenyxe: Some of these are private but also used in the derived class. Maybe we can just make all of them protected, they aren't going to pollute the context of the class a lot more.

Another point is that I don't really see `MergedVariantMongoReader`'s implementation as really derived from `AccessionedVariantMongoReader`. I think using a [Template Method](https://en.wikipedia.org/wiki/Template_method_pattern) that implements `ItemStreamReader`, then making the 2 concrete readers extend it, would improve the code structure.

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/AccessionedVariantMongoReader.java
```diff
@@ -169,25 +171,31 @@ public void open(ExecutionContext executionContext) throws ItemStreamException {
...
-            }
+            addToVariants(variants, contig, start, rs, reference, alternate, sourceEntry);
         }
         return new ArrayList<>(variants.values());
     }
 
-    private long calculateEnd(String reference, String alternate, long start) {
+    protected void addToVariants(Map<String, Variant> variants, String contig, long start, long rs, String reference,
+                               String alternate, VariantSourceEntry sourceEntry) {
+        String variantId = (contig + "_" + start + "_" + reference + "_" + alternate).toUpperCase();
```
cyenyxe: This is a one-liner, but extracting it to a separate method would allow to simplify the `addToVariants` interface and make it more obvious, as in:

```java
protected void addToVariants(Map<String, Variant> variants, String variantId, VariantSourceEntry sourceEntry)
```

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/AccessionedVariantMongoReader.java
```diff
@@ -201,8 +209,8 @@ private VariantSourceEntry buildVariantSourceEntry(String study, String sequence
         return sourceEntry;
     }
 
-    private String buildId(long rs) {
-        return RS_PREFIX + Objects.toString(rs);
+    protected String buildId(long rs) {
```
cyenyxe: Rename to `buildRsId` or similar if the variant ID construction is extracted to a separate method.

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/AccessionedVariantMongoReader.java
```diff
@@ -169,25 +171,31 @@ public void open(ExecutionContext executionContext) throws ItemStreamException {
...
+        } else {
+            long end = calculateEnd(reference, alternate, start);
+            Variant variant = new Variant(contig, start, end, reference, alternate);
+            variant.setMainId(buildId(rs));
+            variant.addSourceEntry(sourceEntry);
+            variants.put(variantId, variant);
+        }
+    }
+
+    protected long calculateEnd(String reference, String alternate, long start) {
```
cyenyxe: I don't think this method needs to be protected because it is not accessed from a derived class.

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/parameters/InputParameters.java
```diff
@@ -47,6 +49,7 @@ public JobParameters toJobParameters() {
                 .addString("assemblyReportUrl", assemblyReportUrl)
                 .addLong("buildNumber", buildNumber)
                 .addString("outputVcf", outputVcf)
+                .addString("outputVcfMerged", outputVcfMerged)
```
cyenyxe: Or we could make it a bit easier to use by deriving the merged and deprecated file names from the `outputVcf` parameter.

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/MergedVariantContextWriter.java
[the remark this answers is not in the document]
andresfsilva: >Current RS ID equivalent to a given variant

?

From the proposed ones I like
> RS ID that currently represents the variant

eva-accession-release/src/test/java/uk/ac/ebi/eva/accession/release/configuration/CreateMergedReleaseStepConfigurationTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+    private void assertStepExecutesAndCompletes() {
+        JobExecution jobExecution = jobLauncherTestUtils.launchStep(RELEASE_MAPPED_MERGED_VARIANTS_STEP);
+        assertEquals(BatchStatus.COMPLETED, jobExecution.getStatus());
+    }
+
+    @Test
+    public void variantsWritten() throws Exception {
+        assertStepExecutesAndCompletes();
+        long numVariantsInRelease = FileUtils.countNonCommentLines(new FileInputStream(inputParameters.getOutputVcfMerged()));
```
cyenyxe: Line too long.

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/MergedVariantMongoReader.java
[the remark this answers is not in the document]
andresfsilva: I like `CURR` or `ACTIVE` and between those two probably `CURR`

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/AccessionedVariantMongoReader.java
jmmut: no, the `start` and `ref` and `alt` are used in the rest of the addToVariants method (the `else` part)

jmmut: unless you want to split the id by underscore to retrieve those fields againg, which I think is a bad idea

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/AccessionedVariantMongoReader.java
jmmut: the part about the aggregation fits the template pattern, but the method `buildVariantSourceEntry` is used in both implementations, and the Merged one only adds an extra variable, so the template pattern doesn't fit very well for that part. Can you think of a better structure (using template pattern) than:
```
///////////// Abstract parent class
protected VariantSourceEntry buildVariantSourceEntry( ... ) {
    new VariantSourceEntry ...
    sourceEntry.addAttribute ...
    // more attributes
    implUpdateVariantSourceEntry(sourceEntry) // overrided
    return sourceEntry;
}

////////////// Implementation class
add variants() {
    // complex method with lots of variables (only way to refactor is to create a class similar to Variant to hold them, which is pointless, or make them fields, and change them for each iteration)
    for ( ... ) {
        buildVariantSourceEntry( ..., **extraParameterMergedInto**)
    }
}
buildVariantSourceEntry( ... , **extraParameterMergedInto**) {
    this.extraParameterMergedInto = extraParameterMergedInto // only way I can think of. ugly
    super.buildVariantSourceEntry(...)
}
implUpdateVariantSourceEntry(sourceEntry){
    sourceEntry.addAttribute(extraParameterMergedInto) // where do we get this parameter? a field?
}
```
or do you prefer mixing both patterns and do the template pattern for the aggregation, and leave it as it is now (simple parent reuse) for the `buildVariantSourceEntry` part?

jmmut: also, the fields that I left private, are not the same. "asm", "study", "contig", "start", "type"; in one class they refer to DbsnpClusteredVariantEntity fields, and the other to the equivalent fields in DbsnpClusteredVariantInactiveEntity. should I still reuse them?

cyenyxe: > in one class they refer to DbsnpClusteredVariantEntity fields, and the other to the equivalent fields in DbsnpClusteredVariantInactiveEntity. should I still reuse them?

The entities have the same internal structure, but one is simply flagged as inactive. I think we should reuse them.

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/MergedVariantContextWriter.java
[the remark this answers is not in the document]
cyenyxe: It is not an equivalency relationship, but a name for that variant.

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/VariantMongoAggregationReader.java
```diff
@@ -0,0 +1,176 @@
...
+    abstract List<Bson> buildAggregation();
+
+    @Override
+    public List<Variant> read() throws Exception, UnexpectedInputException, ParseException, NonTransientResourceException {
+        return cursor.hasNext() ? getVariants(cursor.next()) : null;
+    }
+
+    abstract List<Variant> getVariants(Document clusteredVariant);
+
+    protected VariantSourceEntry buildVariantSourceEntry(String study, Long mergedInto, String sequenceOntology,
```
cyenyxe: When 2 methods have exactly the same signature, the parameter that changes is usually added at the end.

PR #101, opened 2019-02-08
EVA-1436 Only write to FASTA file if more than 1 line is downloaded

cyenyxe:
When the ENA API does not return any results, the line count will be zero.

The previous code would still try to write to the FASTA file, then display a misleading message when `grep` couldn't find the string.

### DIFF
--- a/eva-accession-import-automation/create_fasta_from_assembly_report.sh
+++ b/eva-accession-import-automation/create_fasta_from_assembly_report.sh
@@ -40,7 +40,7 @@ do
     matches=`grep -c "${accession}" ${output_folder}/${assembly_accession}.fa`
     if [ $matches -gt 1 ]
     then
-        echo WARNING: Sequence ${genbank_contig} found more than once in the output FASTA file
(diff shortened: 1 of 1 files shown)


eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -38,8 +40,8 @@ do
     matches=`grep -c "${accession}" ${output_folder}/${assembly_accession}.fa`
     if [ $matches -gt 1 ]
     then
-        echo WARNING: Sequence ${genbank_contig} found more than once in the output FASTA file
-        exit_code=1
+        echo Sequence ${genbank_contig} found more than once in the output FASTA file
+        exit_code=2
```
jmmut: so if a sequence is found more than once it overwrites the exit_code=1? we really can't do anything else

cyenyxe: Yes, any non-zero exit code should be considered a potential error by the top-level automation process and will require manual investigation.

PR #102, opened 2019-02-11
EVA-1436 Prevent duplicate sequences in the output FASTA 

cyenyxe:
By checking if a sequence is already in the FASTA file before writing to it, we make sure there won't be any duplicates. Otherwise the FASTA can't be indexed.

Temporary contig files are also now written to the output folder, instead of the one the script is run from!

### DIFF
--- a/eva-accession-import-automation/create_fasta_from_assembly_report.sh
+++ b/eva-accession-import-automation/create_fasta_from_assembly_report.sh
@@ -14,8 +14,9 @@ output_folder=$3
 
 exit_code=0
 
-# To ensure the FASTA file exists before running `grep -c`, otherwise it will always fail
+# To ensure the file exists before running `grep -c`, otherwise it will always fail
 touch ${output_folder}/${assembly_accession}.fa
+touch ${output_folder}/written_contigs.txt
 
 for genbank_contig in `grep -v -e "^#" ${assembly_report} | cut -f5`;
 do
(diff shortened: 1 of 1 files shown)


eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -14,36 +14,37 @@ output_folder=$3
...
         exit_code=1
         continue
     fi
 
-    # Check that an accession is present no more than once in the output FASTA file, otherwise it 
-    # means there are unexpected aliases in the assembly report, which need to be checked
-    accession=`head -n 1 ${genbank_contig} | cut -f1 -d' ' | cut -f1 -d'.' | cut -f2 -d'>'`
+    # Write the sequence associated with an accession to a FASTA file only once
+    accession=`head -n 1 ${output_folder}/${genbank_contig} | cut -f1 -d' ' | cut -f1 -d'.' | cut -f2 -d'>'`
     matches=`grep -c "${accession}" ${output_folder}/${assembly_accession}.fa`
```
jmmut: when I heard about writing only once, I expected another file with just the written contigs instead of the whole sequence, but ok. we can change that if it causes too much delay. I mean, it could mean grepping a fasta 1000s of times

cyenyxe: Tested it with carrot (ca. 4841 sequences, 352 MB) and it took less than 10 seconds to run to completion, without counting the download time, which seems to be the biggest bottleneck.

eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -14,36 +14,38 @@ output_folder=$3
...
-    then
-        echo Sequence ${genbank_contig} found more than once in the output FASTA file
-        exit_code=2
-    fi
+    # Delete temporary contig file
+    rm ${output_folder}/${genbank_contig}
 done
 
 exit $exit_code
-
```
cyenyxe: Delete the `${output_folder}/written_contigs.txt` file before exiting?

PR #103, opened 2019-02-13
EVA-1433 - Use full genbank equivalents file instead of just assembly related ones

sundarvenkata-EBI:
This PR still uses the genbank equivalent text file from the EBI cluster file system instead of a database table since the memory usage was not prohibitive (around 1GB - and this will be run as a bsub job in conjunction with the main java pipeline which will need more than that anyway)

### DIFF
--- a/eva-accession-import-automation/generate_custom_assembly_report.py
+++ b/eva-accession-import-automation/generate_custom_assembly_report.py
@@ -2,8 +2,8 @@ from get_assembly_report_url import *
 import data_ops
 import os
 import pandas
-import urllib
 import sys
+import urllib.request
 from __init__ import *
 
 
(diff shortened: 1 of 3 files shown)


eva-accession-import-automation/generate_custom_assembly_report.py
```diff
@@ -108,15 +109,14 @@ def update_non_equivalent_genbank_accessions_in_assembly_report(assembly_report_
...
-        equivalent_genbank_accession = \
-            genbank_equivalents_dataframe[genbank_equivalents_dataframe["rs_acc"] == accession]["gb_acc"]
-        if len(equivalent_genbank_accession) == 0:
-            logger.warn("Could not find equivalent GenBank accession for RefSeq accession " + accession)
-        else:
-            equivalent_genbank_accession = equivalent_genbank_accession.values[0]
+        try:
+            equivalent_genbank_accession = \
+                genbank_equivalents_dataframe.loc[accession]["gb_acc"]
             assembly_report_dataframe.loc[assembly_report_dataframe["RefSeq-Accn"] == accession, "GenBank-Accn"] = \
```
andresfsilva: The script is failing when the GenBank contig appears more than once in the equivalences file provided by ncbi.

```GCF_000003025.5	GCA_000003025.4	AF034253.1	NC_000845.1```
```GCF_000003025.6	GCA_000003025.6	AF034253.1	NC_000845.1```

PR #99, opened 2019-02-06
EVA-1438 - Automate the triggering and checking of a dbSNP assembly import

sundarvenkata-EBI:
I have not been able to execute the scripts yet. But I am submitting this PR for initial review.

PR #105, opened 2019-02-15
EVA-1436 create FASTA: add retry logic for wget

jmmut:
with a previous versions I found a silent error where a contig was downloaded partially. I expect wget returns an error exit code in that case.

**Note**: if a contig is not available at all, I kept the behaviour of logging it and continue. the script only stops if there was something downloaded and wget returned non-zero exit code

### DIFF
--- a/eva-accession-import-automation/create_fasta_from_assembly_report.sh
+++ b/eva-accession-import-automation/create_fasta_from_assembly_report.sh
@@ -54,13 +54,6 @@ do
         times_wget_failed=$(($times_wget_failed + 1))
     done
 
-    if [ $times_wget_failed -eq $max_allowed_attempts ]
-    then
-        echo Could not download ${genbank_contig}. FASTA file is left incomplete.
-        exit_code=1
-        break
-    fi
-
     # If a file has more than one line, then it is concatenated into the full assembly FASTA file
     # (empty sequences can't be indexed)
     lines=`head -n 2 ${output_folder}/${genbank_contig} | wc -l`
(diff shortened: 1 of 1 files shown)


eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -22,9 +22,24 @@ for genbank_contig in `grep -v -e "^#" ${assembly_report} | cut -f5`;
...
+        # Download each GenBank accession in the assembly report from ENA into a separate file
+        # Delete the accession prefix from the header line
+        wget -q -O - "https://www.ebi.ac.uk/ena/browser/api/fasta/${genbank_contig}" | sed 's/ENA|.*|//g' > ${output_folder}/${genbank_contig}
+        whole_pipe_result=$?
+        if [ $whole_pipe_result -eq 0 ]
+        then
+            # it was correctly downloaded
+            break
+        fi
+        echo Download for ${genbank_contig} failed. Retrying...
```
andresfsilva: What do you think about only printing the retrying message the first time it has to retry? Otherwise we will get a lot of repetitive messages. We will have this retrying messages a lot for some species specially in the cases when the contigs belong to a WGS as those are not available through the API.

So I would suggest having this
```
CM000815.5
CM000816.5
Download for CM000816.5 failed. Retrying...
FASTA sequence not available for CM000816.5
CM000817.5
Download for CM000817.5 failed. Retrying...
FASTA sequence not available for CM000817.5
```
instead of this
```
CM000815.5
CM000816.5
Download for CM000816.5 failed. Retrying...
Download for CM000816.5 failed. Retrying...
Download for CM000816.5 failed. Retrying...
Download for CM000816.5 failed. Retrying...
Download for CM000816.5 failed. Retrying...
FASTA sequence not available for CM000816.5
CM000817.5
Download for CM000817.5 failed. Retrying...
Download for CM000817.5 failed. Retrying...
Download for CM000817.5 failed. Retrying...
Download for CM000817.5 failed. Retrying...
Download for CM000817.5 failed. Retrying...
FASTA sequence not available for CM000817.5
```

At least while we find out if there is a way to know the contig belong to a WGS in which case we wouldn't need to try downloading it through the API but going straight to the FTP

jmmut: makes sense. if the contig was successfully downloaded at the second of third time, it won't print "sequence not available"

jmmut:
I'm waiting for #99 to be merged to solve a simple conflict with it

PR #110, opened 2019-02-20
EVA-1455 RS IDs deprecation pipelines

andresfsilva:
New module `eva-accession-deprecate`

Clustered variants that were declustered during the import and are not used in the main collection will be deprecated. The deprecation includes writing in the operation collection and removing from the declustered collection


pom.xml
```diff
@@ -95,7 +96,7 @@
     <dependencies>
         <dependency>
             <!--https://docs.spring.io/spring-boot/docs/1.5.3.RELEASE/reference/html/configuration-metadata.html#
-            configuration-metadata-annotation-processor-->
+            test-metadata-annotation-processor-->
```
sundarvenkata-EBI: The new URL doesn't seem to work!

eva-accession-deprecate/src/main/java/uk/ac/ebi/eva/accession/deprecate/configuration/ClusteredVariantsToDeprecateReaderConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+
+import uk.ac.ebi.eva.accession.core.configuration.MongoConfiguration;
+import uk.ac.ebi.eva.accession.deprecate.io.ClusteredVariantsToDeprecateReader;
+
+@Configuration
+@Import({MongoConfiguration.class})
+public class ClusteredVariantsToDeprecateReaderConfiguration {
```
sundarvenkata-EBI: Suggest renaming to DeprecableClusteredVariantsReaderConfiguration. See https://en.wiktionary.org/wiki/deprecable

eva-accession-deprecate/src/main/java/uk/ac/ebi/eva/accession/deprecate/io/ClusteredVariantsToDeprecateReader.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Read all ClusteredVariants in the dbsnpClusteredVariantEntityDeclustered collection that are not in the main
+ * collection dbsnpClusteredVariantEntity
+ *
+ * ClusteredVariantsToDeprecateReader = dbsnpClusteredVariantEntityDeclustered - dbsnpClusteredVariantEntity
+ */
+public class ClusteredVariantsToDeprecateReader implements ItemStreamReader<DbsnpClusteredVariantEntity> {
```
sundarvenkata-EBI: see https://github.com/EBIvariation/eva-accession/pull/110/files#r258481548

eva-accession-deprecate/src/test/java/uk/ac/ebi/eva/accession/deprecate/io/ClusteredVariantsToDeprecateReaderTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Before
+    public void setUp() {
+        executionContext = new ExecutionContext();
+        reader = new ClusteredVariantsToDeprecateReader(mongoClient, TEST_DB);
+    }
+
+    @Test
+    public void ReadVariantsToDeprecate() {
+        List<DbsnpClusteredVariantEntity> variants = readIntoList();
```
sundarvenkata-EBI: Just as a matter of principle, I suggest we should test at least 2 values when reading lists of results.

eva-accession-deprecate/src/main/java/uk/ac/ebi/eva/accession/deprecate/configuration/ClusteredVariantsToDeprecateReaderConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Bean
+    @StepScope
+    ClusteredVariantsToDeprecateReader clusteredVariantsToDeprecateReader(MongoClient mongoClient,
+                                                                          MongoProperties mongoProperties) {
```
sundarvenkata-EBI: Since we are using only the database name, why pass in the full MongoProperties object? Let's just pass in the database name.

eva-accession-deprecate/src/main/java/uk/ac/ebi/eva/accession/deprecate/io/ClusteredVariantsToDeprecateReader.java
```diff
@@ -0,0 +1,135 @@
...
+        String contig = declusteredVariant.getString(CONTIG_FIELD);
+        long start = declusteredVariant.getLong(START_FIELD);
+        String type = declusteredVariant.getString(TYPE_FIELD);
+        long accession = declusteredVariant.getLong(ACCESSION_KEY);
+        int version = declusteredVariant.getInteger(VERSION_KEY);
+        LocalDateTime createdDate = declusteredVariant.getDate("createdDate")
+                                                      .toInstant().atZone(ZoneId.systemDefault()).toLocalDateTime();
+
+        return new DbsnpClusteredVariantEntity(accession, id, assembly, taxonomy, contig, start,
+                                               VariantType.valueOf(type), false, createdDate, version);
```
sundarvenkata-EBI: Validated flag is explicitly set to false here. Even though the validated flag for a variant in the deprecated collection might not be very important, let's just preserve what was in the source for data integrity reasons.

eva-accession-deprecate/src/test/java/uk/ac/ebi/eva/accession/deprecate/io/ClusteredVariantsToDeprecateReaderTest.java
```diff
@@ -0,0 +1,97 @@
...
+        assertEquals(1, variants.size());
+        assertEquals(ID_1, variants.get(0).getId());
+        assertEquals(ACCESSION_1, variants.get(0).getAccession());
+    }
+
+    private List<DbsnpClusteredVariantEntity> readIntoList() {
+        reader.open(executionContext);
```
sundarvenkata-EBI: Shall we move reader open and close to setup and teardown methods?

eva-accession-deprecate/src/test/java/uk/ac/ebi/eva/accession/deprecate/test/configuration/MongoTestConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.data.mongodb.config.EnableMongoAuditing;
+import org.springframework.data.mongodb.repository.config.EnableMongoRepositories;
+
+@Configuration
+@EntityScan(basePackages = {"uk.ac.ebi.eva.accession.core.persistence"})
+@EnableMongoRepositories(basePackages = {
+        "uk.ac.ebi.eva.accession.core.persistence",
+        "uk.ac.ebi.ampt2d.commons.accession.persistence.mongodb.repository"})
+@EnableMongoAuditing
```
sundarvenkata-EBI: I am curious as to why we need auditing for test data for which we are using fake data anyway.

jmmut: I think this was for filling in the date timestamps

eva-accession-deprecate/src/main/java/uk/ac/ebi/eva/accession/deprecate/io/ClusteredVariantsToDeprecateReader.java
jmmut: in fact, maybe we can use the converter from mongoTemplate, but it uses the old mongo api with DBObjects instead of Documents.
```
        MongoConverter converter = mongoTemplate.getConverter();
        DbsnpClusteredVariantEntity read = converter.read(DbsnpClusteredVariantEntity.class, new BasicDBObject());
```
so either keep this as it is, or do the aggregation with the old api and the automatic converter

eva-accession-deprecate/src/main/java/uk/ac/ebi/eva/accession/deprecate/configuration/ClusteredVariantsToDeprecateReaderConfiguration.java
andresfsilva: As we are using `MongoConfiguration` we already have a Bean for `MongoProperties`. It will read from the properties file the parameters with the prefix `spring.data.mongodb` and the database name will be there. If we pass the String for the database we will need to define a Bean for it and read from the properties file.

I would prefer use the `MongoProperties` object

eva-accession-deprecate/src/main/java/uk/ac/ebi/eva/accession/deprecate/io/DeprecationWriter.java
```diff
@@ -0,0 +1,89 @@
...
+        BulkOperations bulkOperations = mongoTemplate.bulkOps(BulkOperations.BulkMode.UNORDERED,
+                                                              DbsnpClusteredVariantEntity.class,
+                                                              DBSNP_CLUSTERED_VARIANT_DECLUSTERED_COLLECTION_NAME);
+        Query query = new Query();
+        query.addCriteria(Criteria.where(ID_FIELD).in(getIds(deprecableClusteredVatiants)));
+        bulkOperations.remove(query);
+        bulkOperations.execute();
+    }
+
+    private List<String> getIds(List<? extends DbsnpClusteredVariantEntity> deprecableClusteredVatiants) {
```
sundarvenkata-EBI: correct typo "variants" instead of "vatiants"

eva-accession-deprecate/src/main/java/uk/ac/ebi/eva/accession/deprecate/io/DeprecationWriter.java
```diff
@@ -0,0 +1,89 @@
...
+        insertDeprecateOperation(deprecableClusteredVariants);
+    }
+
+    private void removeDeprecableClusteredVariants(
+            List<? extends DbsnpClusteredVariantEntity> deprecableClusteredVatiants) {
+        BulkOperations bulkOperations = mongoTemplate.bulkOps(BulkOperations.BulkMode.UNORDERED,
+                                                              DbsnpClusteredVariantEntity.class,
+                                                              DBSNP_CLUSTERED_VARIANT_DECLUSTERED_COLLECTION_NAME);
+        Query query = new Query();
+        query.addCriteria(Criteria.where(ID_FIELD).in(getIds(deprecableClusteredVatiants)));
```
sundarvenkata-EBI: I am not sure if there are limits on how many elements the in clause can contain. Have you checked if  the deprecable clustered variant population low enough to be contained in a list and not to cause any issues query-wise?

sundarvenkata-EBI: Never mind, I realized that the list of items is governed by chunk size.

eva-accession-deprecate/src/test/java/uk/ac/ebi/eva/accession/deprecate/configuration/DeprecateClusteredVariantsStepConfigurationTest.java
```diff
@@ -0,0 +1,106 @@
...
+        "/test-data/dbsnpClusteredVariantEntity.json",
+        "/test-data/dbsnpClusteredVariantEntityDeclustered.json"})
+@TestPropertySource("classpath:application.properties")
+public class DeprecateClusteredVariantsStepConfigurationTest {
+
+    private static final String TEST_DB = "test-db";
+
+    private static final String DBSNP_CLUSTERED_VARIANT_ENTITY_DECLUSTERED = "dbsnpClusteredVariantEntityDeclustered";
+
+    private static final long EXPECTED_DECLUSTERED_VARIANTS_DECLUSTERED = 2;
```
sundarvenkata-EBI: Suggest renaming to EXPECTED_VARIANTS_TO_BE_FULLY_DECLUSTERED

eva-accession-deprecate/src/main/java/uk/ac/ebi/eva/accession/deprecate/io/DeprecationWriter.java
```diff
@@ -0,0 +1,89 @@
...
+    private MongoTemplate mongoTemplate;
+
+    public DeprecationWriter(MongoTemplate mongoTemplate) {
+        this.mongoTemplate = mongoTemplate;
+    }
+
+    @Override
+    public void write(List<? extends DbsnpClusteredVariantEntity> deprecableClusteredVariants) throws Exception {
+        removeDeprecableClusteredVariants(deprecableClusteredVariants);
+        insertDeprecateOperation(deprecableClusteredVariants);
```
jmmut: after thinking again about this, I think I prefer to insert first. As we don't have transactions, if something fails with the current order, we might loose track of RSs that should be deprecated and they won't appear in the main RS collection nor in the RS operations collection, only in the SS operations.

If something fails while inserting first, we can end up with repeated RS operations, which is recoverable.

eva-accession-deprecate/src/main/java/uk/ac/ebi/eva/accession/deprecate/io/DeprecationWriter.java
```diff
@@ -0,0 +1,89 @@
...
+                                                              DbsnpClusteredVariantEntity.class,
+                                                              DBSNP_CLUSTERED_VARIANT_DECLUSTERED_COLLECTION_NAME);
+        Query query = new Query();
+        query.addCriteria(Criteria.where(ID_FIELD).in(getIds(deprecableClusteredVatiants)));
+        bulkOperations.remove(query);
+        bulkOperations.execute();
+    }
+
+    private List<String> getIds(List<? extends DbsnpClusteredVariantEntity> deprecableClusteredVariants) {
+        return deprecableClusteredVariants.stream().flatMap(d -> Stream.of(d.getId())).collect(Collectors.toList());
```
jmmut: why not `deprecableClusteredVariants.stream().map(AccessionedDocument::getId).collect(Collectors.toList())` ?

eva-accession-deprecate/src/main/java/uk/ac/ebi/eva/accession/deprecate/io/DeprecationWriter.java
cyenyxe: Agree with writing first. This is also how moving between disk partitions work, first copying then deleting, to avoid losing any information.

jmmut: so one thing we can do is catch any exception  in this function and log the list of accessions, in case we have to recover the chunk manually

PR #111, opened 2019-02-27
EVA-1436 Include WGS sequences in FASTA

andresfsilva:
The ENA API feature about redirecting to the FTP when retrieving a contig that belongs to a WGS sequence is still not deployed in production. Once it is the URL must be changed to use `www` instead of `wwwdev`

### DIFF
--- a/eva-accession-import-automation/create_fasta_from_assembly_report.sh
+++ b/eva-accession-import-automation/create_fasta_from_assembly_report.sh
@@ -39,15 +39,15 @@ do
     set -o pipefail
     times_wget_failed=0
     max_allowed_attempts=5
-	
-	# Check if the contig belong to a WGS sequence that was already downloaded
-	already_downloaded=`grep -c "${genbank_contig}" ${output_folder}/written_contigs.txt`
-	if [ $already_downloaded -eq 1 ]
-	then
+
+    # Check if the contig belong was already downloaded
+    already_downloaded=`grep -c "${genbank_contig}" ${output_folder}/written_contigs.txt`
+    if [ $already_downloaded -eq 1 ]
+    then
         echo 'contig already in the FASTA file'
         continue
-	fi
-	
+    fi
+
     while [ $times_wget_failed -lt $max_allowed_attempts ]
     do
         # Download each GenBank accession in the assembly report from ENA into a separate file
--- a/eva-accession-import-automation/create_fasta_from_assembly_report.sh
+++ b/eva-accession-import-automation/create_fasta_from_assembly_report.sh
@@ -41,10 +41,10 @@ do
     max_allowed_attempts=5
 
     # Check if the contig has already been downloaded
-    already_downloaded=`grep -c "${genbank_contig}" ${output_folder}/written_contigs.txt`
+    already_downloaded=`grep -F -c "${genbank_contig}" ${output_folder}/written_contigs.txt`
     if [ $already_downloaded -eq 1 ]
     then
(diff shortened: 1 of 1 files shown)


eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -39,11 +39,20 @@ do
     set -o pipefail
     times_wget_failed=0
     max_allowed_attempts=5
+
+    # Check if the contig belong was already downloaded
```
cyenyxe:
```suggestion
    # Check if the contig has already been downloaded
```

eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -39,11 +39,20 @@ do
...
+        echo 'contig already in the FASTA file'
+        continue
+    fi
+
     while [ $times_wget_failed -lt $max_allowed_attempts ]
     do
         # Download each GenBank accession in the assembly report from ENA into a separate file
         # Delete the accession prefix from the header line
-        wget -q -O - "https://www.ebi.ac.uk/ena/browser/api/fasta/${genbank_contig}" | sed 's/ENA|.*|//g' > ${output_folder}/${genbank_contig}
+        wget -q -O - "https://wwwdev.ebi.ac.uk/ena/browser/api/fasta/${genbank_contig}" | sed 's/ENA|.*|//g' > ${output_folder}/${genbank_contig}
```
cyenyxe: ENA said today this is now in production.

eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -39,11 +39,20 @@ do
     set -o pipefail
     times_wget_failed=0
     max_allowed_attempts=5
+
+    # Check if the contig belong was already downloaded
+    already_downloaded=`grep -c "${genbank_contig}" ${output_folder}/written_contigs.txt`
+    if [ $already_downloaded -eq 1 ]
+    then
+        echo 'contig already in the FASTA file'
```
cyenyxe:
```suggestion
        echo Contig ${genbank_contig} is already present in the FASTA file and doesn't need to be downloaded again.
```

eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -79,8 +88,23 @@ do
             matches=`grep -m 1 -c "${genbank_contig}" ${output_folder}/written_contigs.txt`
             if [ $matches -eq 0 ]
             then
+                # Check if is a gws. If it is the file will be compressed
```
cyenyxe:
```suggestion
                # If the downloade file contains a WGS, it will be compressed
```

eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -39,11 +39,20 @@ do
     set -o pipefail
     times_wget_failed=0
     max_allowed_attempts=5
+
+    # Check if the contig belong was already downloaded
+    already_downloaded=`grep -c "${genbank_contig}" ${output_folder}/written_contigs.txt`
```
sundarvenkata-EBI: Since the contig has a dot in it, we might have to tell grep explicitly to not treat that dot as a regular expression (use the -F switch). I guess this won't be an issue going with the assumption that genbank contigs are of fixed length but we have had bad history with assumptions!

PR #113, opened 2019-02-28
EVA-1458 Log message only once per contig

andresfsilva:
Log only once per contig when a contig and a chromosome are not in the same line of the assembly report

### DIFF
--- a/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/ContigReplacerProcessor.java
+++ b/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/ContigReplacerProcessor.java
@@ -59,9 +59,9 @@ public class ContigReplacerProcessor implements ItemProcessor<SubSnpNoHgvs, SubS
                             + assemblyAccession + "' correct?");
         }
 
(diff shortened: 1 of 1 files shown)


eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/ContigReplacerProcessor.java
```diff
@@ -31,33 +34,39 @@
...
+                    "Neither contig '" + contigName + "' nor chromosome '"
                             + subSnpNoHgvs.getChromosome()
                             + "' were found in the assembly report! Is the assembly accession '"
                             + assemblyAccession + "' correct?");
         }
 
-        if (chromosomePresentInAssemblyReport
+        if (!processedContigs.contains(contigName)
+                && chromosomePresentInAssemblyReport
                 && contigPresentInAssemblyReport
```
jmmut: really pedantic, but putting the check for contains after both boolean checks will save lots of strings hash creation.

sundarvenkata-EBI: +1

PR #118, opened 2019-03-05
EVA-1436 Fix grep after removing accessions prefix

andresfsilva:


eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -95,15 +95,15 @@ do
                     # Uncompress file
                     mv ${output_folder}/${genbank_contig} ${output_folder}/${genbank_contig}.gz
                     gunzip ${output_folder}/${genbank_contig}.gz
-                    # Delete the accessions prefix
-                    sed -i 's/ENA|.*|//g' ${output_folder}/${genbank_contig}
                 fi
+                # Delete the accessions prefix
+                sed -i 's/ENA|.*|//g' ${output_folder}/${genbank_contig}
```
sundarvenkata-EBI: Going into ultra-paranoid mode now that this script has been arising from the dead many times: 
I am suggesting a guard against white-spaces in the accessions line. This way, any brain-damaged input with spaces can be converted to the desired output.

image

PR #119, opened 2019-03-08
EVA-1462 Import zebrafish_7955 into accessioning service

jmmut:
allow importing no-variants, i.e. variants where all the alleles are the same as the reference allele, for instance rs40632972:

ref: ACACACACACACAC
alleles: (AC)7/7


eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/model/SubSnpNoHgvs.java
```diff
@@ -268,8 +269,13 @@ public String getReferenceInForwardStrand() {
     public List<String> getAlternateAllelesInForwardStrand() {
         String referenceAllele = getReferenceInForwardStrand();
         List<String> alleles = getAllelesInForwardStrand();
+        List<String> allelesWithoutReference = new ArrayList<>(alleles);
 
-        return alleles.stream().filter(allele -> !allele.equals(referenceAllele)).collect(Collectors.toList());
+        // even if the reference allele is present several times, remove it only once. if we removed all the
+        // occurrences of the reference and there were no other alleles, the pipeline would skip this variant.
+        allelesWithoutReference.remove(referenceAllele);
```
cyenyxe: It may sound a bit paranoid, but I would prefer to remove all the occurrences, then add one back. It is less efficient, but who knows if there are variants like `C/C/C` out there...

jmmut: I thought it would be better to allow the writer to detect collisions and issue merges as needed. do you still think it's better to remove all but one?

cyenyxe: Yes, I think it is not very useful to create a merge event due to poor formatting/curation, when it is so easy to avoid in this particular scenario.

jmmut: ok. will do. it's not exactly "remove all the occurrences, then add one back", though:

|alleles | desired alleles|
|---|---|
|R/A     | A|
|R/A/T   | A/T|
|R/R     | R|
|R/R/R   | R|
|R/R/A/T | A/T|

PR #127, opened 2019-04-05
EVA-1273 Link new SS IDs to GenBank contig accession whenever possible, otherwise RefSeq accession

andresfsilva:


eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/processors/ContigReplacerProcessor.java
```diff
@@ -0,0 +1,98 @@
...
+    @Override
+    public IVariant process(IVariant variant) throws Exception {
+        String contigName = variant.getChromosome();
+        ContigSynonyms contigSynonyms = contigMapping.getContigSynonyms(contigName);
+        boolean contigPresentInAssemblyReport = contigSynonyms != null;
+
+        if (!contigPresentInAssemblyReport) {
+            if (!processedContigs.contains(contigName)) {
+                logger.warn("Contig '" + variant.getChromosome() + "' was not found in the assembly report!");
+                processedContigs.add(contigName);
```
jmmut: what do others think about adding a method to log the total count of missing contigs? (also, split processedContigs in missingContigs and nonEquivalentContigs and log both):

```

    @AfterStep
        // declared as fields: Set<String> missingContigs, nonEquivalentContigs;
        if (!missingContigs.isEmpty()) {
            logger.warn("{} contigs were not present in the assembly report", missingContigs.size());
        }
        if (!nonEquivalentContigs.isEmpty()) {
            logger.warn("{} contigs were not replaced by INSDC names because the sequence was not identical",
                        nonEquivalentContigs.size());
        }
        return stepExecution.getExitStatus();
    }
```
(I just learnt that `@AfterStep` exists)

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/configuration/VariantProcessorConfiguration.java
```diff
@@ -66,4 +68,14 @@ VariantProcessor variantProcessor(InputParameters inputParameters) {
     ExcludeStructuralVariantsProcessor excludeStructuralVariantsProcessor() {
         return new ExcludeStructuralVariantsProcessor();
     }
+
+    @Bean
+    ContigReplacerProcessor contigReplacerProcessor(ContigMapping contigMapping) {
```
jmmut: you are not using this processor in the `compositeProcessor.setDelegates(...)` call

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/processors/ContigReplacerProcessor.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * The conversion will be performed only if the 'relationship' column of the assembly report is '='
+     */
+    private IVariant replaceContigWithSynonym(IVariant variant, ContigSynonyms contigSynonyms) {
+        if (contigSynonyms.isIdenticalGenBankAndRefSeq()) {
+            Variant newVariant = new Variant(getContigSynonym(contigSynonyms), variant.getStart(), variant.getEnd(),
+                                             variant.getReference(), variant.getAlternate());
+            Collection<VariantSourceEntry> sourceEntries = variant.getSourceEntries().stream()
+                                                                  .map(VariantSourceEntry.class::cast)
+                                                                  .collect(Collectors.toList());
```
jmmut: ugh. This is so ugly but I can't think of a better thing to do. we don't currently need the source entries, but the factory will add them. we could:

- make the whole step to use `Variant` instead of `IVariant`.
- and/or use  `new VcfReader(new CoordinatesVcfLineMapper(), new File(inputParameters.getOutputVcf()));` and assert here that the sourceEntries is empty. after all we only care about chr, start, ref, alt. the project, taxonomy and assembly is provided as input parameter.
- or not copying the sourceEntries (dangerous or misleading to reuse this class?)
- or keep it as it is now?

any other idea?

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/processors/ContigReplacerProcessor.java
andresfsilva: We would need to register the `ContigReplacerProcessor` as a listener in the Step for the annotation to work because we are using a composite processor. Discussed offline with @jmmut and agreed not to do it.

https://stackoverflow.com/questions/39697529/spring-batch-use-before-and-after-step-methods-within-a-compositeitemwriter

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/steps/processors/ContigReplacerProcessorTest.java
```diff
@@ -0,0 +1,81 @@
...
+    public void ContigNotFoundInAssemblyReport() throws Exception {
+        IVariant variant = new Variant("chr", 1, 1, "A", "T");
+        assertEquals("chr", processor.process(variant).getChromosome());
+    }
+
+    @Test
+    public void NoGenbankContigUseRefseq() throws Exception {
+        IVariant variant = new Variant("chr4", 1, 1, "A", "T");
+        assertEquals("NW_003763687.1", processor.process(variant).getChromosome());
+    }
```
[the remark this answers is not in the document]
andresfsilva: I wasn't sure about this case but ended up doing it this way because the ticket said `GenBank otherwise RefSeq` but I agree with @jmmut

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/processors/ContigReplacerProcessor.java
andresfsilva: Although I don't like it a lot I would go for doing the cast (as it is now), my second option would be using `Variant` instead of `IVariant` and the cast wouldn't be needed.

I Will leave it with the cast unless someone would prefer another option.

eva-accession-pipeline/src/test/resources/input-files/assembly-report/assembly_report.txt
```diff
@@ -0,0 +1,7 @@
+# mocking of an assembly report
+# Sequence-Name	Sequence-Role	Assigned-Molecule	Assigned-Molecule-Location/Type	GenBank-Accn	Relationship	RefSeq-Accn	Assembly-Unit	Sequence-Length	UCSC-style-name
+chr1	assembled-molecule	1	Chromosome	CM000093.4	=	NW_003763476.1	Primary Assembly	196202544	chr1
+chr2	assembled-molecule	2	Chromosome	CM000094.4	=	NW_003763901.1	Primary Assembly	149560735	chr2
+chr3	assembled-molecule	3	Chromosome	CM000095.4	<>	NW_003763686.1	Primary Assembly	111302122	chr3
+chr4	assembled-molecule	4	Chromosome	na	=	NW_003763687.1	Primary Assembly	111302123	chr4
+chr5	assembled-molecule	5	Chromosome	na	<>	NW_003763688.1	Primary Assembly	111302124	chr5
```
sundarvenkata-EBI: May be it is not plausible, but just for completeness, could you add tests for "na <> na" and "Genbank <> na"?

PR #130, opened 2019-04-11
EVA-1492 Fix RS release failure on big species (active and merged IDs)

jmmut:


cyenyxe:
Waiting for some test runs to finish before merging.

PR #137, opened 2019-06-14
Barebones dbsnp json import pipeline impl

108krohan:
### What's in the change? 
- Barebones impl of the import pipeline. 
- Following fields are brought in as of now: refsnp_id, assembly_name, and start, end, ref, and alt fields. 

### Resources
- for mappings used in Jackson parser, please see [rsjson-demo.py](https://github.com/ncbi/dbsnp/blob/master/tutorials/rsjson_demo.py)

### Caveats
- I should not need to disable the configurations that are not related to my job flow.  (Will attempt to self-rectify in the weekend)
- May need to update the reader for parsing for bulk Json.BZ files. Need to update read of Json file and test on a sample subset of those.
- As of now, the connection doesn't close (while I have been able to get over that [OptimisticLockingFailureException](https://stackoverflow.com/questions/17680863/attempt-to-update-step-execution-id-1-with-wrong-version-2-where-current-vers)), the data base connection doesn't terminate on it's own after the object is persisted.

### Testing
Checked the database table Variants.  Variant object persists!

> db.variants.count()
1
> db.variants.find();
{ "_id" : "NC_000008.11_19956017_A_G", "chr" : "NC_000008.11", "start" : NumberLong(19956017), "_at" : { "chunkIds" : [ "NC_000008.11_19956_1k", "NC_000008.11_1995_10k" ] }, "alt" : "G", "end" : NumberLong(19956018), "files" : [ {  } ], "hgvs" : [ { "type" : "genomic", "name" : "NC_000008.11:g.19956017A>G" } ], "len" : 1, "ref" : "A", "type" : "SNV" }


108krohan:
The build might fail for lack of beans/configurations creations because they are commented out as of now. Would resolve in a future fix.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpJsonItemReader.java
```diff
@@ -0,0 +1,95 @@
...
+    private InputParameters parameters;
+
+    public DbsnpJsonItemReader(InputParameters parameters) {
+        this.parameters = parameters;
+    }
+
+    @Override
+    public DbsnpJson read() throws Exception {
+        JsonFactory jsonFactory = new JsonFactory();
+        JsonParser jp = jsonFactory.createJsonParser(new File(parameters.getDbsnp()));
```
sundarvenkata-EBI: This method seems to be deprecated since version 2.2. Use the other method suggested [here](https://fasterxml.github.io/jackson-core/javadoc/2.8/com/fasterxml/jackson/core/JsonFactory.html#createJsonParser(byte[]))

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpJsonItemReader.java
```diff
@@ -0,0 +1,95 @@
...
+        JsonNode jsonNode = jp.readValueAsTree();
+        return parseDbsnpJsonPlacements(jsonNode);
+    }
+
+    /**
+     * Parses the JSON as per https://github.com/ncbi/dbsnp/blob/master/tutorials/hadoop_json_placement.py
+     * @param JsonNode jsonRootNode
+     * @return parsed DbsnpJson object
+     */
+    private DbsnpJson parseDbsnpJsonPlacements(JsonNode jsonRootNode) throws IOException {
```
sundarvenkata-EBI: This does not seem to deal with multi-allelic variants (ex: A/T,G). Either 

1. DbsnpJson has to be modified to accommodate multiple alternates or 
2. This function should return multiple DbsnpJson objects. 

I think the former is easier from an implementation standpoint. @cyenyxe Any thoughts?

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpJsonItemReader.java
```diff
@@ -0,0 +1,95 @@
...
+
+            boolean isPtlp = alleleInfo.path("is_ptlp").asBoolean();
+            JsonNode assemblyInfo = alleleInfo.path("placement_annot").path("seq_id_traits_by_assembly");
+            if(assemblyInfo.size() == 0 || !isPtlp) {
+                continue;
+            }
+
+            dbsnpJson.setProjectAccession(assemblyInfo.get(0).path("assembly_name").asText());
+
+            for (JsonNode alleleAndHgvs : alleleInfo.path("alleles")) {
```
sundarvenkata-EBI: Have you consulted Ensembl's code to check if this is how they parse the variants? A good exercise at this time would be to run Ensembl's code and this pipeline in parallel for a representative sample of variants (that should include multi-allelic variants) and verify if the results match.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpJsonItemReader.java
```diff
@@ -0,0 +1,95 @@
...
+                JsonNode spdi = alleleAndHgvs.path("allele").path("spdi");
+                String insertedSequence = spdi.path("inserted_sequence").asText();
+                String deletedSequence = spdi.path("deleted_sequence").asText();
+                if(!insertedSequence.equals(deletedSequence)) {
+                    dbsnpJson.setReferenceAllele(deletedSequence);
+                    dbsnpJson.setAlternateAllele(insertedSequence);
+                    dbsnpJson.setStart(spdi.path("position").asLong());
+                    dbsnpJson.setContig(spdi.path("seq_id").asText());
```
sundarvenkata-EBI: Have you consulted Ensembl's code to check if this is how they parse the variants? A good exercise at this time would be to run Ensembl's code and this pipeline in parallel for a representative sample of variants (that should include multi-allelic variants) and verify if the results match.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpJsonItemReader.java
108krohan: Nice catch! Thanks!

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpJsonItemReader.java
108krohan: I would need to setup their code on my machine for that. Going through their Perl parsing logic as of now.

So if I understand correctly, actionable steps: 
1. Isolate 50 JSON fields (check if it has multi-allelic variants defined, if not add more JSON attrs)
2. The above would require to modify the reader to accept a JSON array, instead of single JSON. 
3. Setup Ensembl's code on my machine and run the two pipelines in parallel.

sundarvenkata-EBI: You don't have to add more JSON attributes. You just have to look for multiple occurrences of the "allele" attribute within **primary_snapshot_data -> placements_with_allele ->alleles**. See 
[rs1413947121.zip](https://github.com/EBIvariation/eva-accession/files/3300318/rs1413947121.zip) and the [NCBI link](https://www.ncbi.nlm.nih.gov/snp/rs1413947121)

108krohan: So, because an iVariant can have only 1 alternate [variation-commons/iVariant](https://github.com/EBIvariation/variation-commons/blob/fda0d8bbf44f029ab7ef1f756f31964b3bc9e2c5/variation-commons-core/src/main/java/uk/ac/ebi/eva/commons/core/models/IVariant.java)
What that means is to call itemProcessor for every loop iteration.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/io/DbsnpJsonItemReader.java
cyenyxe: If `DbsnpJson` supports multiple alternates, they will need to be split later on when we construct the final objects to be serialized. But it is true that it may be easier to deal with that split at that point instead of during the reading, where the input is quite convoluted.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/model/DbsnpJson.java
```diff
@@ -0,0 +1,74 @@
+package uk.ac.ebi.eva.accession.dbsnp.model;
+
+import org.springframework.batch.core.configuration.annotation.StepScope;
+
+/**
+ * Dbsnp object model to store key fields from Dbsnp Json
+ */
+@StepScope
```
cyenyxe: You don't need to annotate this class, `@StepScope` only affects `@Bean`.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/model/DbsnpJson.java
```diff
@@ -0,0 +1,74 @@
+package uk.ac.ebi.eva.accession.dbsnp.model;
+
+import org.springframework.batch.core.configuration.annotation.StepScope;
+
+/**
+ * Dbsnp object model to store key fields from Dbsnp Json
```
cyenyxe:
```suggestion
 * Core fields of a variant extracted from a dbSNP JSON file.
```

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/parameters/InputParameters.java
```diff
@@ -22,6 +22,8 @@
 
     private String fasta;
 
+    private String dbsnp;
```
cyenyxe: This parameter name is unclear.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/persistence/DbsnpVariant.java
```diff
@@ -0,0 +1,33 @@
...
+import uk.ac.ebi.eva.commons.core.models.AbstractVariant;
+import uk.ac.ebi.eva.commons.core.models.IVariantSourceEntry;
+import uk.ac.ebi.eva.commons.core.models.VariantStatistics;
+import uk.ac.ebi.eva.commons.core.models.pipeline.VariantSourceEntry;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+public class DbsnpVariant extends AbstractVariant {
```
cyenyxe: I don't think you need this class, `eva-accession` already contains types for submitted and clustered variant, and another one to group them together.

eva-accession-import/src/main/resources/application.properties
```diff
@@ -1,38 +1,34 @@
...
-parameters.assemblyName=
-parameters.assemblyReportUrl=
-parameters.taxonomyAccession=
-parameters.chunkSize=100
-#parameters.pageSize=
+spring.batch.job.names=IMPORT_DBSNP_JSON_VARIANTS_JOB
+
+# dbsnp
+parameters.dbsnp=/Users/rohnkum/casesafe/acsn/refsnp-sample.json
+parameters.dbsnpAccession=12345
```
cyenyxe: Is this the build number? If so, rename it appropriately.

eva-accession-import/src/main/resources/application.properties
```diff
@@ -1,38 +1,34 @@
...
-parameters.assemblyAccession=
-parameters.assemblyName=
-parameters.assemblyReportUrl=
-parameters.taxonomyAccession=
-parameters.chunkSize=100
-#parameters.pageSize=
+spring.batch.job.names=IMPORT_DBSNP_JSON_VARIANTS_JOB
+
+# dbsnp
+parameters.dbsnp=/Users/rohnkum/casesafe/acsn/refsnp-sample.json
```
cyenyxe: No paths in properties files please. Also, we normally name the input file `parameters.input` in other pipelines.

eva-accession-import/pom.xml
```diff
@@ -51,7 +51,29 @@
             <artifactId>nosqlunit-mongodb</artifactId>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
```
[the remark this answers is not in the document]
108krohan: Ok now I understand what you mean by making the pipeline iteratively. I had it sidewards. I thought to bring in only a couple of fields only at first. But what it also means is to first get the readers write, then move over to the processors and the writers. And then maybe the validations accordingly.

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/model/DbsnpJson.java
108krohan: Corrections noted (and those from above as well)!

eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/parameters/InputParameters.java
108krohan: dbsnp_source

108krohan: How about that for param name^^^

PR #139, opened 2019-06-19
EVA-1370 - Re-run accessioning in Embassy Cloud for a given submission

sundarvenkata-EBI:


eva-accession-pipeline/automation/run_accessioning_in_embassy_cloud.py
```diff
@@ -0,0 +1,112 @@
...
+import click
+import glob
+import json
+import subprocess
+import os
+import configparser
+import sys
+from __init__ import logger
+
+sys.path.append("../../eva-accession-import-automation")
```
jmmut: I guess there isn't a cleaner way to do this, right?

sundarvenkata-EBI: We can't use relative path imports when module name has hyphens.

PR #140, opened 2019-06-19
EVA-1561 - Check existing contigs in PROD for a given assembly for bug impact

sundarvenkata-EBI:


eva-accession-import-automation/validate_imported_contigs_for_assembly.py
```diff
@@ -0,0 +1,132 @@
...
+    get_contigs_start_mismatch_cmd = "psql -A -t -h {metahost} -U {metauser} -d {metadb} -c " \
+                                     "\"select distinct contig_name from {contig_chr_mismatch_table} " \
+                                     "where table_name like '%{assembly_md5}%'" \
+                                     " and schema_name like '%{taxonomy_accession}%'\" -P pager=off " \
+                                     "| grep {assembly_report_path} -f - |".format(**config) + final_formatting_cmd
+    get_contigs_start_match_cmd = "psql -A -t -h {metahost} -U {metauser} -d {metadb} -c " \
+                                  "\"select distinct contig_name from {contig_chr_match_table} " \
+                                  "where table_name like '%{assembly_md5}%'" \
+                                  " and schema_name like '%{taxonomy_accession}%'\" -P pager=off " \
+                                  "| grep {assembly_report_path} -f - | grep -v assembled-molecule |".format(**config) \
```
jmmut: why removing the assembled-molecules?

eva-accession-import-automation/validate_imported_contigs_for_assembly.py
```diff
@@ -0,0 +1,132 @@
...
+    config.update(get_args_from_assembly_properties_file(assembly_properties_file))
+    config["assembly_report_path"] = config["parameters.assemblyReportUrl"].split("file:")[-1]
+    config["assembly_md5"] = hashlib.md5(config["parameters.assemblyName"].encode("utf-8")).hexdigest()
+    # We need these because string interpolation won't work if there is a dot in the key
+    config["taxonomy_accession"] = config["parameters.taxonomyAccession"]
+
+
+    config["contig_chr_mismatch_table"] = "dbsnp_ensembl_species.dbsnp_species_with_contig_chromosome_start_mismatch"
+    config["contig_chr_match_table"] = "dbsnp_ensembl_species.dbsnp_species_with_contig_chromosome_start_match"
+    final_formatting_cmd = r"cut -d$'\t' -f5 | sort | uniq | paste -s - | sed s/$'\t'/\",\"/g | sed s/$/\"/g | sed s/^/\"/g"
```
jmmut: f5 is the genbank column right? is there any way to mention that within the code without comments?

eva-accession-import-automation/validate_imported_contigs_for_assembly.py
sundarvenkata-EBI: We are taking out the assembled-molecule lines because we are looking for contigs where contig_start = chr_start but are not actually in the assembled-molecule entries. 

Given that RS2 is a RefSeq contig in chr1 in dbSNP with **contig_start = chr_start** and the following in the assembly report:
chr1 assembled-molecule CM1 = RS1
RS2  scaffold NONCM1 = RS2 

The candidate that we should look for is RS2 but not RS1 because RS1 would have definitely been replaced with CM1 by the older version of the pipeline but RS2 won't be.

jmmut: oh, I see, this is why sheep and another species give false positives. ok

eva-accession-import-automation/validate_imported_contigs_for_assembly.py
```diff
@@ -0,0 +1,132 @@
...
+                                 "--password {3} --authenticationDatabase=admin " \
+                                 "{4} --eval 'db.dbsnpSubmittedVariantEntity.findOne({{\"seq\": \"{5}\", " \
+                                 "\"contig\": {{$in: [{6}]}}}})'"
+
+    mismatch_contig_set = run_command_with_output("Get contigs with start mismatch against chromosome:",
+                                                  get_contigs_start_mismatch_cmd)[1].strip()
+    match_contig_set = run_command_with_output("Get contigs with start match against chromosome:",
+                                               get_contigs_start_match_cmd)[1].strip()
+
+    if mismatch_contig_set != '""':
```
cyenyxe: The idiom in Python does not need the comparison:
```suggestion
    if mismatch_contig_set:
```

eva-accession-import-automation/validate_imported_contigs_for_assembly.py
```diff
@@ -0,0 +1,132 @@
...
+                          shell=True) as process:
+        for line in process.stdout:
+            process_output += line
+        errors = os.linesep.join(process.stderr.readlines())
+    if process.returncode != 0:
+        logger.error(command_description + " failed!" + os.linesep + errors)
+        raise subprocess.CalledProcessError(process.returncode, process.args)
+    else:
+        logger.info(command_description + " completed successfully")
+    return process.returncode, process_output
```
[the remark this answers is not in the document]
sundarvenkata-EBI: The exit code won't be anything other than zero so I will take the exit code out from the return value

eva-accession-import-automation/validate_imported_contigs_for_assembly.py
sundarvenkata-EBI: This won't work because what we are comparing against (the output from the sed command) is not an empty string but a literal set of double quotes.

PR #141, opened 2019-06-21
EVA-1562 - Generate one assembly report per assembly

sundarvenkata-EBI:


eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -33,6 +33,8 @@ exit_code=0
 touch ${output_folder}/${species}_custom.fa
 touch ${output_folder}/written_contigs.txt
 
+grep '^>' ${output_folder}/${species}_custom.fa | cut -d' ' -f1 | sed 's/>//g' >> ${output_folder}/written_contigs.txt
```
cyenyxe: Appending to an existing FASTA file could add duplicates to `written_contigs.txt`. This will be a problem at least when running [this comparison](https://github.com/EBIvariation/eva-accession/pull/141/files#diff-a64fd7a82abcaeecd117c07fa7e5aaa9R111).

But if we decide to keep it (because it is true that it could save quite a bunch of time) and update the rest of the code, then I would comment what we are trying to do here:
```suggestion
# If using a pre-existing FASTA file, identify the sequences already downloaded to avoid doing it again
grep '^>' ${output_folder}/${species}_custom.fa | cut -d' ' -f1 | sed 's/>//g' >> ${output_folder}/written_contigs.txt
```

sundarvenkata-EBI: written_contigs.txt lives only for the duration of the program (see https://github.com/EBIvariation/eva-accession/pull/141/files#diff-a64fd7a82abcaeecd117c07fa7e5aaa9L175). so it won't exist when the program starts.

PR #143, opened 2019-06-23
Import dbSNP JSON Variant: reader implementation

108krohan:
## Change
- New dbSNP JSON import module with reader implementation.
- Read from BZ2 compressed file on the fly
- Jackson parsing


eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/io/DbsnpJsonItemProcessor.java
```diff
@@ -0,0 +1,77 @@
...
+     * Parses the JSON into a list of dbSNP submitted variants
+     * @see <a https://github.com/Ensembl/ensembl-variation/blob/master/scripts/import/dbSNP_v2/load_dbsnp.pl#L331>
+     *     ensembl-import-dbsnp-v2</a>
+     * @param jsonRootNode root node of a dbSNP JSON line item
+     * @return list of dbSNP submitted variants
+     */
+    private List<DbsnpJson> parseDbsnpJsonPlacements(JsonNode jsonRootNode) throws IOException {
+        List<DbsnpJson> dbsnpJsons = new ArrayList<>();
+        DbsnpJson dbsnpJson = new DbsnpJson();
+        dbsnpJson.setClusteredVariantAccession(jsonRootNode.path("refsnp_id").asLong());
```
sundarvenkata-EBI: Is asLong null resistant or will this statement explode if there are no REFSNP_IDs?

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/io/DbsnpJsonItemProcessor.java
```diff
@@ -0,0 +1,77 @@
...
+        JsonNode infoNode = jsonRootNode.path("primary_snapshot_data").path("placements_with_allele");
+
+        for(JsonNode alleleInfo : infoNode) {
+            boolean isPtlp = alleleInfo.path("is_ptlp").asBoolean();
+            JsonNode assemblyInfo = alleleInfo.path("placement_annot").path("seq_id_traits_by_assembly");
+            if(assemblyInfo.size() == 0 || !isPtlp) {
+                continue;
+            }
+            dbsnpJson.setProjectAccession(assemblyInfo.get(0).path("assembly_name").asText());
```
sundarvenkata-EBI: Why is assembly name set as the project accession?

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/model/DbsnpJson.java
```diff
@@ -0,0 +1,126 @@
...
+    private long start;
+
+    private String referenceAllele;
+
+    private String alternateAllele;
+
+    private Long clusteredVariantAccession;
+
+    public DbsnpJson() {
+        taxonomyAccession = 9606;
```
sundarvenkata-EBI: I am a bit torn on doing this in the code. I understand that this code is primarily going to be for human import only but I was wondering if this is better supplied as an input parameter. @cyenyxe Any thoughts?

sundarvenkata-EBI: Also, we should take in the assemblyAccession (GCA equivalent of the GCF assembly) as an input parameter 
image

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/model/DbsnpJson.java
```diff
@@ -0,0 +1,126 @@
...
+import org.springframework.batch.core.configuration.annotation.StepScope;
+import uk.ac.ebi.eva.accession.core.ISubmittedVariant;
+
+import java.time.LocalDateTime;
+
+/**
+ * Core fields of a variant extracted from a dbSNP JSON file.
+ */
+@StepScope
+public class DbsnpJson implements ISubmittedVariant {
```
sundarvenkata-EBI: This should also extend AccessionedDocument<ISubmittedVariant, Long> to account for the SubSNP accession itself (which should also be captured btw. see **primary_snapshot_data->support** attribute with **id->type** as **subsnp**)

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/io/DbsnpJsonItemProcessor.java
108krohan: Not to persist any value at all? 

suggest to replace with ```Long.valueOf(jsonRootNode.path("refsnp_id"))```

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/model/DbsnpJson.java
108krohan: suggestion: Derive from Input Param > hard-code

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/Application.java
```diff
@@ -0,0 +1,35 @@
...
+package uk.ac.ebi.eva.accession.dbsnp2;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.autoconfigure.batch.BatchAutoConfiguration;
+import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
+import org.springframework.boot.autoconfigure.orm.jpa.HibernateJpaAutoConfiguration;
+import org.springframework.context.ConfigurableApplicationContext;
+
+@SpringBootApplication(exclude = {
```
cyenyxe: Again, this should not be necessary and doing it complicates the application configuration. I will try to debug what is going on today.

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/io/BzipLazyResource.java
```diff
@@ -0,0 +1,55 @@
...
+        super(path);
+    }
+
+    @Override
+    public InputStream getInputStream() throws IOException {
+        BufferedInputStream bis = new BufferedInputStream(super.getInputStream());
+        try {
+            return new CompressorStreamFactory().createCompressorInputStream(bis);
+        } catch (CompressorException compressorException) {
+            throw new IOException("ZipException: Not in BZ2 format");
```
cyenyxe: Using names that are apparently one exception but have another is misleading, I recommend a more generic message:

```suggestion
            throw new IOException("The input file is not compressed in bzip2 format");
```

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/io/DbsnpJsonItemProcessor.java
cyenyxe: The benefit of Mongo is that you don't need to persist information that is not available, i.e. null, which allows to save space. In the import of dbSNP 1.0 we already avoid persisting the RS ID when it is not present; the code should do it automatically as soon as the input is converted to the appropriate models.

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/io/DbsnpJsonItemReader.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.springframework.batch.item.file.FlatFileItemReader;
+import org.springframework.batch.item.file.mapping.JsonLineMapper;
+import org.springframework.core.io.Resource;
+
+import java.io.File;
+import java.util.Map;
+
+/**
+ * Reader of dbSNP JSON flat file. The file should be zipped.
```
cyenyxe:
```suggestion
 * Reader of dbSNP JSON flat file. The file should be compressed in bzip2 format.
```

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/model/DbsnpJson.java
cyenyxe: > I am a bit torn on doing this in the code. I understand that this code is primarily going to be for human import only but I was wondering if this is better supplied as an input parameter. @cyenyxe Any thoughts?

I cringed a bit at first, but being honest this code is not "primarily going to be", it is "only going to be" for human imports for several years. Given that, I would make this a constant instead of a variable.

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/model/DbsnpJson.java
```diff
@@ -0,0 +1,126 @@
...
+
+import org.springframework.batch.core.configuration.annotation.StepScope;
+import uk.ac.ebi.eva.accession.core.ISubmittedVariant;
+
+import java.time.LocalDateTime;
+
+/**
+ * Core fields of a variant extracted from a dbSNP JSON file.
+ */
+@StepScope
```
cyenyxe: This is not necessary, the scope applies when declaring something to be used as a `@Bean`. This class lifecycle is not managed by Spring.

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/model/DbsnpJson.java
cyenyxe: I don't see the need for this class at all. Input objects should be converted to [SubmittedVariant](https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariant.java), which means that one can simply call the appropriate constructor from the processor.

For serialization, a [DbsnpVariantsWrapper](https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/persistence/DbsnpVariantsWrapper.java) grouping both the submitted and clustered variant **database entities** is constructed and used as input for database serialization.

Please check the module `eva-accession-import` for reference on how the pipeline currently works, on we can go through it during one of the calls.

eva-accession-import-dbsnp2/src/main/resources/application.properties
```diff
@@ -0,0 +1,19 @@
+spring.batch.job.names=IMPORT_DBSNP_JSON_VARIANTS_JOB
+
+# dbSNP JSON source
+parameters.dbsnpSource=
```
cyenyxe: The rest of our pipelines use `parameters.input` to specify the primary input file.

eva-accession-import-dbsnp2/src/test/java/uk/ac/ebi/eva/accession/dbsnp2/io/DbsnpJsonItemReaderTest.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(SpringRunner.class)
+@ContextConfiguration(classes = {BatchTestConfiguration.class})
+@TestPropertySource({"classpath:application.properties"})
+public class DbsnpJsonItemReaderTest {
+
+    private String sourceFile = StringUtils.EMPTY;
```
cyenyxe: Using a library to create an empty string is overkill, specially when the value of the attribute is set in every test.
```suggestion
    private String sourceFile;
```

eva-accession-import-dbsnp2/src/test/java/uk/ac/ebi/eva/accession/dbsnp2/test/BatchTestConfiguration.java
```diff
@@ -0,0 +1,69 @@
...
+import uk.ac.ebi.eva.accession.dbsnp2.configuration.ImportDbsnpJsonVariantsStepConfiguration;
+import uk.ac.ebi.eva.accession.dbsnp2.configuration.ImportDbsnpJsonFlowConfiguration;
+import uk.ac.ebi.eva.accession.dbsnp2.configuration.InputParametersConfiguration;
+import uk.ac.ebi.eva.accession.dbsnp2.runner.DbsnpJsonImportVariantsJobLauncherCommandLineRunner;
+import uk.ac.ebi.eva.commons.batch.job.JobExecutionApplicationListener;
+
+@EnableAutoConfiguration
+@Import({MongoConfiguration.class,
+        ImportDbsnpJsonVariantsJobConfiguration.class,
+        ImportDbsnpJsonVariantsStepConfiguration.class,
```
cyenyxe: The configuration classes should be able to deal with the nesting via `@Import` themselves.

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/io/DbsnpJsonItemProcessor.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class DbsnpJsonItemProcessor implements ItemProcessor<Map<String, Object>, List<DbsnpJson>> {
+
+    private ObjectMapper mapper = new ObjectMapper();
+
+    public List<DbsnpJson> process(Map<String, Object> jsonItem) throws Exception {
+        JsonNode jsonNode = mapper.valueToTree(jsonItem);
```
108krohan: return jsonNode

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/model/DbsnpJson.java
108krohan: please reopen if any follow-ups

sundarvenkata-EBI: For assembly accession, my proposal would be to take the following two as input parameters:

1) Equivalent GCA accession for the GCF (see screenshot in the above comment). This will map to the referenceSequenceAccession member.
2) The GCF accession itself 

For a given pipeline run with the 2 given parameters above, the code should only use data in the "alleles" section under those documents where **primary_snapshot_data->placements_with_allele->placement_annot->seq_id_traits_by_assembly->assembly_accession** equals the GCF accession from the input parameter. @cyenyxe What do you think?

PR #150, opened 2019-07-11
EVA-1576 Add the replacement chromosome in the metadata section in the accession report

jmmut:
I had to add an extra step `BuildReportTasklet`. To write all the contigs in the metadata section we have to use temporary files to write the contigs and the variants as they are processed, and at the end concatenate those files into the final VCF, and remove the temporary files. It wasn't possible to recover from a failed job and still build a correct VCF from the AccessionReportWriter because the only place to do this final concatenation would be the "close" method and it is also called when there's an error (and the temporary files shouldn't be deleted in that case).

Also, the conversion from IVariant to ISubmittedVariant is not a processor anymore, because the AccessionReportWriter needs the original chromosome, which is available only in IVariant, not ISubmittedVariant.


eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/buildReport/BuildReportTaskletTest.java
```diff
@@ -0,0 +1,146 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.file.Files;
+import java.util.Arrays;
+
+import static org.junit.Assert.*;
```
jmmut: remove wildcard

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningService.java
```diff
@@ -59,24 +61,31 @@ public SubmittedVariantAccessioningService(SubmittedVariantMonotonicAccessioning
...
-                variantsNotInDbsnp);
-        return joinLists(submittedVariants, dbsnpVariants);
+        if (variantsNotInDbsnp.isEmpty()) {
+            // check this special case because mongo bulk inserts don't allow inserting empty lists
+            // (accession-commons BasicMongoDbAccessionedCustomRepositoryImpl.insert would need to change)
+            return dbsnpVariants;
+        } else {
+            List<AccessionWrapper<ISubmittedVariant, String, Long>> submittedVariants = accessioningService.getOrCreate(
+                    variantsNotInDbsnp);
+            return joinLists(submittedVariants, dbsnpVariants);
```
sundarvenkata-EBI: Are the above changes related to EVA-1576?

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningService.java
```diff
@@ -59,24 +61,31 @@ public SubmittedVariantAccessioningService(SubmittedVariantMonotonicAccessioning
...
-        return allVariants.stream().filter(variant -> !contains(vatiantsToDelete, variant))
+                                                           variantsToDelete) {
+        Set<String> hashesToDelete = variantsToDelete.stream()
+                                                     .map(AccessionWrapper::getHash)
+                                                     .collect(Collectors.toSet());
+
+        return allVariants.stream()
+                          .map(variant -> Pair.of(accessioningServiceDbsnp.getHash(variant), variant))
+                          .filter(pair -> !hashesToDelete.contains(pair.getFirst()))
+                          .map(Pair::getSecond)
```
sundarvenkata-EBI: Is this to address [this issue](https://ebi-eva.slack.com/archives/C5A6MLDAR/p1562166268149400)?

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningServiceTest.java
```diff
@@ -218,6 +218,41 @@ public void getOrCreateAccessionsInBothRepositories() throws AccessionCouldNotBe
...
+    }
+
+    @UsingDataSet(locations = {"/test-data/dbsnpSubmittedVariantEntity.json"})
+    @Test
+    public void getOrCreateDbsnpEquivalentVariants() throws AccessionCouldNotBeGeneratedException {
+        SubmittedVariant identicalVariantPresentInDbsnp = new SubmittedVariant("GCA_000009999.3", 9999, PROJECT_DBSNP,
+                                                                               "21", 20849999, "", "GG", null);
+
+        List<AccessionWrapper<ISubmittedVariant, String, Long>> generatedAccessions = service.getOrCreate(
+                Arrays.asList(identicalVariantPresentInDbsnp));
```
sundarvenkata-EBI: Shouldn't identicalVariantPresentInDbsnp be used in the previous function?

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriter.java
```diff
@@ -84,31 +111,65 @@ public void setAccessionPrefix(String accessionPrefix) {
...
-            boolean append = true;
-            this.fileWriter = new BufferedWriter(new FileWriter(this.output, append));
-            if (!isHeaderAlreadyWritten) {
-                writeHeader();
-                executionContext.put(IS_HEADER_WRITTEN_KEY, IS_HEADER_WRITTEN_VALUE);
+            if (isHeaderAlreadyWritten) {
+                // we are resuming a job
+                if (contigsOutput.exists() && variantsOutput.exists()) {
+                    loadContigMappingFromTemporaryFile(contigsOutput);
+                    boolean append = true;
```
sundarvenkata-EBI: Since the append variable's value isn't being changed, should this be set here?

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriter.java
```diff
@@ -117,31 +178,81 @@ public void update(ExecutionContext executionContext) throws ItemStreamException
...
+    private String getOriginalChromosome(IVariant variant) {
+        Set<String> originalChromosomes = variant.getSourceEntries()
+                                                 .stream()
+                                                 .map(se -> se.getAttributes().get(
+                                                         ContigToGenbankReplacerProcessor.ORIGINAL_CHROMOSOME))
+                                                 .collect(Collectors.toSet());
+
+        if (originalChromosomes.size() != 1) {
+            throw new IllegalStateException(
+                    "Bug detected: Can not provide the original chromosome of a variant because there should be "
```
sundarvenkata-EBI: Suggest to re-phrase to "Bug detected: Multiple original chromosomes were found to be associated with the same variant" and then add variant info, the associated original chromosomes and the replacement used.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriter.java
```diff
@@ -117,31 +178,81 @@ public void update(ExecutionContext executionContext) throws ItemStreamException
...
-            writeVariant(variant.getAccession(), variant.getData());
+        variantsWriter.flush();
+    }
+
+    private void updateChromosomeMappings(List<? extends IVariant> originalVariantsWithReplacedContigs)
+            throws IOException {
+        for (IVariant variantWithContig : originalVariantsWithReplacedContigs) {
+            String originalChromosome = getOriginalChromosome(variantWithContig);
+
+            String previousContig = inputContigsToInsdc.put(originalChromosome, variantWithContig.getChromosome());
```
sundarvenkata-EBI: Suggest collecting variantWithContig.getChromosome() in a variable currentContig and use that subsequently. This makes the rest of the code intuitive.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningService.java
jmmut: tangentially related, yes. I needed this to make the other changes. it was discussed offline not to create a ticket just for this
> -- question: if a variant with hash1 is present in dbsnp and we are asked to accession a variant that has the same hash1 but some other non-identifying fields different, should we
> 1) have 2 (similar but not identical) documents, one in SVE and the other in dbsnpSVE that have the same hash
> 2) return the dbsnp accession
> ?
> because right now we are doing option 1)  basically since the start of EVA SS issuing.
> dbsnp accessions are returned only if all the object matches exactly, including taxonomy, the flags for mismatches and validation, and the rs; which is unlikely to happen
> 
> -- how can the hash be the same if the study is different?
> -- well, ok I didn't think about that. so in practice it won't happen, but we have some PRJEBs in the dbsnp side that submitters could ask to add analysis to

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningService.java
jmmut: yes

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningServiceTest.java
jmmut: what do you mean? in `@Test getOrCreateDbsnpIdenticalVariants()` ? the same variant is used there. here it has "identical" in the name to contrast with the other variable "equivalent"

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriter.java
jmmut: another value is used later in this function, and I think 

```
this.contigsWriter = new BufferedWriter(new FileWriter(this.contigsOutput, true));
[...]
this.contigsWriter = new BufferedWriter(new FileWriter(this.contigsOutput, false));
```
is more obscure than
```
boolean append = true;
this.contigsWriter = new BufferedWriter(new FileWriter(this.contigsOutput, append));
[...]
boolean append = false;
this.contigsWriter = new BufferedWriter(new FileWriter(this.contigsOutput, append));
```

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriter.java
```diff
@@ -117,31 +178,81 @@ public void update(ExecutionContext executionContext) throws ItemStreamException
 
     public void close() throws ItemStreamException {
         try {
-            fileWriter.close();
+            contigsWriter.close();
+            variantsWriter.close();
+            if (!duplicatedInputContigsToInsdc.isEmpty()) {
```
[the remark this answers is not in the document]
jmmut: me neither. I guess I'm ok stopping the job and reevaluate if we face this issue

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/buildReport/BuildReportTasklet.java
```diff
@@ -0,0 +1,122 @@
...
+            throws IOException {
+        writer.write("##fileformat=VCFv4.2");
+        writer.newLine();
+
+        inputContigAndInsdcPairs.forEach((Pair<String, String> inputContigAndInsdc) -> {
+            try {
+                writer.write("##contig=<ID=" + inputContigAndInsdc.getFirst() + ",Description=\""
+                             + inputContigAndInsdc.getSecond() + "\">");
+                writer.newLine();
+            } catch (IOException e) {
```
sundarvenkata-EBI: Why do we handle an exception here but not in the method below? Also why should it be transformed to an RTE?

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriterTest.java
```diff
@@ -164,160 +172,140 @@ public static String getFirstVariantLine(File output) throws IOException {
...
-        AccessionWrapperComparator accessionWrapperComparator = new AccessionWrapperComparator(
-                Collections.singletonList(variant));
-
-        accessionReportWriter.write(Collections.singletonList(accessionWrapper), accessionWrapperComparator);
-
-        assertEquals(String.join("\t", CHROMOSOME_1, Integer.toString(START_1 - 1), ACCESSION_PREFIX + ACCESSION,
-                                 denormalizedReference, denormalizedAlternate,
-                                 ".", ".", "."),
-                     getFirstVariantLine(output));
+        writeVariantWithAccessionHelper(1, "", "A", 1, "G", "AG");
```
sundarvenkata-EBI: Why not use the symbolic variables as in the previous method?

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriterTest.java
```diff
@@ -164,160 +172,140 @@ public static String getFirstVariantLine(File output) throws IOException {
...
-
-        AccessionWrapperComparator accessionWrapperComparator = new AccessionWrapperComparator(
-                Collections.singletonList(variant));
-
-        accessionReportWriter.write(Collections.singletonList(accessionWrapper), accessionWrapperComparator);
-
-        assertEquals(
-                String.join("\t", CHROMOSOME_1, "1", ACCESSION_PREFIX + ACCESSION, "GT", "T", ".", ".", "."),
-                getFirstVariantLine(output));
+        writeVariantWithAccessionHelper(1, "G", "", 1, "GT", "T");
```
sundarvenkata-EBI: Why not use the symbolic variables as in the previous method?

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/SubmittedVariantAccessioningServiceTest.java
```diff
@@ -218,6 +218,41 @@ public void getOrCreateAccessionsInBothRepositories() throws AccessionCouldNotBe
...
 
+    @UsingDataSet(locations = {"/test-data/dbsnpSubmittedVariantEntity.json"})
+    @Test
+    public void getOrCreateDbsnpIdenticalVariants() throws AccessionCouldNotBeGeneratedException {
+        SubmittedVariant variantPresentInDbsnp = new SubmittedVariant("GCA_000009999.3", 9999, PROJECT_DBSNP, "21",
+                                                                      20849999, "", "GG", null);
+
+        List<AccessionWrapper<ISubmittedVariant, String, Long>> generatedAccessions = service.getOrCreate(
+                Collections.singletonList(variantPresentInDbsnp));
+        assertEquals(1, generatedAccessions.size());
```
sundarvenkata-EBI: Would be better to test the accession itself (ACCESSION_DBSNP_1) rather than the number of generated accessions.

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/buildReport/BuildReportTasklet.java
jmmut: lambdas can throw exceptions in java, but classes like Function can't do so:
```
@FunctionalInterface
public interface Function<T, R> {
    R apply(T t);    // <-- no checked exceptions in signature!
    [...]
```

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriterTest.java
jmmut: oversight, and wasn't sure the variables added value. will use them anyway for consistency

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/tasklets/reportCheck/AccessionWrapperComparator.java
```diff
@@ -18,26 +18,31 @@
...
-        this.contigOrder = getContigOrder(variants);
+    /**
+     * @param variants      The contig order will be extracted from this initial variant list
+     */
+    public AccessionWrapperComparator(List<? extends IVariant> variants) {
+        this.contigOrder = getContigOrder(variants.stream().map(IVariant::getChromosome).collect(Collectors.toList()));
     }
 
-    private Map<String, Integer> getContigOrder(List<? extends ISubmittedVariant> variants) {
+    private static Map<String, Integer> getContigOrder(Iterable<String> contigs) {
```
andresfsilva: this doesn't need to be static

eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/io/AccessionReportWriter.java
```diff
@@ -84,31 +111,65 @@ public void setAccessionPrefix(String accessionPrefix) {
...
+        BufferedReader contigReader = new BufferedReader(new FileReader(contigMappingFile));
+        String line;
+        while ((line = contigReader.readLine()) != null) {
+            String[] contigColumns = line.split("\t");
+            if (contigColumns.length != 2) {
+                throw new IllegalStateException("Temporary file " + contigMappingFile.getAbsolutePath()
+                                                + " doesn't have the expected format. Please delete it and "
+                                                + "start a new job.");
+            }
+            inputContigsToInsdc.put(contigColumns[0], contigColumns[1]);
```
andresfsilva: can you extract the `contigColumns[0]` and `contigColumns[1]` to make the code more clear?

jmmut: I did. did you do the last review on outdated code?

andresfsilva: No I reviewed the latest version, maybe there is some confusion because there was a similar comment before but it was in another class

PR #152, opened 2019-07-15
EVA-1581 Delete FASTA index on custom FASTA creation

jmmut:


eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -33,6 +33,9 @@ exit_code=0
 touch ${output_folder}/${species}_custom.fa
 touch ${output_folder}/written_contigs.txt
 
+# remove the FASTA index because it might get outdated (EVA-1581)
+rm ${output_folder}/${species}_custom.fa.fai
```
sundarvenkata-EBI: Please use the "-f" switch so that no error is raised in case of a non-existent fai file.

eva-accession-import-automation/create_fasta_from_assembly_report.sh
```diff
@@ -33,6 +33,9 @@ exit_code=0
 touch ${output_folder}/${species}_custom.fa
 touch ${output_folder}/written_contigs.txt
 
+# remove the FASTA index because it might get outdated (EVA-1581)
+rm -f ${output_folder}/${species}_custom.fa.fai
```
andresfsilva: will we delete the FASTA index everytime this script is executed or should we only do it when new contigs are written to the FASTA file?

sundarvenkata-EBI: this will require a flag to check if at least one contig has been added by the script. To be on the simpler side, I would suggest to remove FAI unconditionally (just in case there were modifications to the FASTA outside of the script).

jmmut: the pipelines will generate the index anyway, and it's just some seconds. I prefer unconditional removal too

andresfsilva: ok, let's leave it as is

PR #155, opened 2019-07-24
EVA-1598 Use ss operations instead of ss collection when querying for the merged file

andresfsilva:
To create the merged file in the RS release a join between `dbsnpClusteredVariantOperationEntity` and `dbsnpSubmittedVariantEntity` was performed but doing that was returning more variants than it should.

Having `RS1 / SS1` and `RS2 / SS2`
Hash for both RS is the same so they are `merged RS2 -> RS1` and a `MERGED` operation is registered
`SS2` is updated with `RS1` and an `UPDATE` operation is registered

With the current query the join would find the `MERGED` operation fir RS1 and would look for the SS IDs associated to it in the `dbsnpSubmittedVariantEntity` but SS1 was no involved in the merged operation

The new query will join with `dbsnpSubmittedVariantOperationEntity` instead and variants will be **filtered out** if:
- Submitted variant operation type is not `UPDATED`.
- Chromosome and start are different in Clustered and Submitted variant operations


eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/MergedVariantMongoReader.java
```diff
@@ -69,11 +69,12 @@ public void open(ExecutionContext executionContext) throws ItemStreamException {
...
     List<Bson> buildAggregation() {
-        Bson matchAssembly = Aggregates.match(Filters.eq(getInactiveField(REFERENCE_ASSEMBLY_FIELD), assemblyAccession));
+        Bson matchAssembly = Aggregates
+                .match(Filters.eq(getInactiveField(REFERENCE_ASSEMBLY_FIELD), assemblyAccession));
         Bson matchMerged = Aggregates.match(Filters.eq(EVENT_TYPE_FIELD, EventType.MERGED.toString()));
         Bson sort = Aggregates.sort(orderBy(ascending(getInactiveField(CONTIG_FIELD), getInactiveField(START_FIELD))));
-        Bson lookup = Aggregates.lookup(DBSNP_SUBMITTED_VARIANT_ENTITY, MERGE_INTO_FIELD,
-                                        CLUSTERED_VARIANT_ACCESSION_FIELD, SS_INFO_FIELD);
+        Bson lookup = Aggregates.lookup(DBSNP_SUBMITTED_VARIANT_OPERATION_ENTITY, ACCESSION_FIELD,
+                                        getInactiveField(CLUSTERED_VARIANT_ACCESSION_FIELD), SS_INFO_FIELD);
```
sundarvenkata-EBI: Have you ensured that the lookup from the SVOE yields appropriate counts (at least for a few species)?

eva-accession-release/src/test/java/uk/ac/ebi/eva/accession/release/io/MergedVariantMongoReaderTest.java
```diff
@@ -162,32 +160,24 @@ public void checkMergedInto() throws Exception {
...
-
-        assertEquals("T", variants.get(ID_2).getReference());
-        assertEquals("G", variants.get(ID_2).getAlternate());
     }
 
     @Test
     public void includeValidatedFlag() throws Exception {
         assertFlagEqualsInAllVariants(CLUSTERED_VARIANT_VALIDATED_KEY, false);
-        assertFlagEqualsInAllVariants(SUBMITTED_VARIANT_VALIDATED_KEY, false);
+        assertFlagEqualsInAllVariants(SUBMITTED_VARIANT_VALIDATED_KEY, true);
```
sundarvenkata-EBI: Just to avoid confusion, could this incidental code change (to support the data) not relevant to the PR changes be avoided by modifying the test data?

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/MergedVariantMongoReader.java
```diff
@@ -87,9 +88,9 @@ private String getInactiveField(String field) {
         Collection<Document> inactiveObjects = (Collection<Document>) mergedVariant.get(INACTIVE_OBJECTS);
         if (inactiveObjects.size() > 1) {
             throw new AssertionError("The class '" + this.getClass().getSimpleName()
-                                     + "' was designed assuming there's only one element in the field "
-                                     + "'" + INACTIVE_OBJECTS + "'. Found " + inactiveObjects.size()
-                                     + " elements in _id=" + mergedVariant.get(ACCESSION_FIELD));
+                                             + "' was designed assuming there's only one element in the field "
```
cyenyxe: Restore indentation.

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/MergedVariantMongoReader.java
```diff
@@ -69,11 +69,12 @@ public void open(ExecutionContext executionContext) throws ItemStreamException {
 
     @Override
     List<Bson> buildAggregation() {
-        Bson matchAssembly = Aggregates.match(Filters.eq(getInactiveField(REFERENCE_ASSEMBLY_FIELD), assemblyAccession));
+        Bson matchAssembly = Aggregates
+                .match(Filters.eq(getInactiveField(REFERENCE_ASSEMBLY_FIELD), assemblyAccession));
```
cyenyxe: Weird, should fit in one line.

eva-accession-release/src/test/java/uk/ac/ebi/eva/accession/release/io/MergedVariantMongoReaderTest.java
```diff
@@ -162,32 +160,24 @@ public void checkMergedInto() throws Exception {
...
         Map<String, Variant> variants = readIntoMap();
-        assertEquals(3, variants.size());
+        assertEquals(EXPECTED_MERGED_VARIANTS, variants.size());
 
         assertEquals("G", variants.get(ID_1_A).getReference());
         assertEquals("A", variants.get(ID_1_A).getAlternate());
 
         assertEquals("G", variants.get(ID_1_T).getReference());
         assertEquals("T", variants.get(ID_1_T).getAlternate());
-
```
cyenyxe: Why not checking a third element for the sake of completion?

eva-accession-release/src/test/java/uk/ac/ebi/eva/accession/release/io/MergedVariantMongoReaderTest.java
```diff
@@ -212,7 +202,38 @@ public void includeAllelesMatchFlag() throws Exception {
...
 
+    /**
+     * This test will use a different reader for assembly GCA_000001215.4 to evaluate this specific scenario:
+     * - 1 Merge operation in dbsnpClusteredVariantOperationEntity (rs881301177 merged into rs80393223)
+     * - 2 Submitted variants in dbsnpSubmittedVariantEntity (ss99056614, ss1986084768) with the same rs80393223
+     * - 1 Update operation in dbsnpSubmittedVariantOperationEntity for ss1986084768 (Original rs881301177 was merged
+     * into rs80393223)
+     * - 1 Merge operation in dbsnpSubmittedVariantOperationEntity for ss1986084768
+     *
+     * Even though the rs80393223 was involved in a merge operation it doesn't mean all of it's associated variants were
```
cyenyxe:
```suggestion
     * Even though the rs80393223 was involved in a merge operation it doesn't mean all of its associated variants were
```

andresfsilva:
> Even though I'm mostly fine with the code, does this change conceptually make sense from a user perspective? If the RS is the one that gets merged into another, won't it be somewhat confusing if we restrict the list of alleles that we return?

@cyenyxe The problem with the previous code is that we would be saying that we merged more variants than we really did. 

If we have two RS IDs each with one SS ID and we merge one RS into the other the merged variants should be just one and we were listing two.

Merged VCF:
```
#CHROM  POS     ID      REF     ALT     QUAL    FILTER  INFO
AE013599.5      7680719 rs881301177     GTATG   G       .       .       CURR=rs80393223;
AE013599.5      7680719 rs881301177     GT      G       .       .       CURR=rs80393223;
```

In this example `rs881301177` was merged into `rs80393223` but the first variant was never linked with the `rs881301177`

cyenyxe:
I see, then yeah, it is definitely relevant. Just take a look at my other mostly minor comments :)

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/MergedVariantMongoReader.java
andresfsilva: I checked some of the species we released in the first round but they all seem to either have the issue or have no merged variants at all

PR #156, opened 2019-07-26
Add processor to replace contig with Genbank accession

108krohan:
### Revision 2

- Resources directory restructure
- Trim assembly report file

#### Commit 
- Add processor to replace Contig name to GenBank accession synonym
   - **Note**: Based on [NCBI Spec as discussed in Schema](https://docs.google.com/spreadsheets/d/1mztm11rHi12eucm9BFMoTFQ5qFVExeLTHm8diClGRNU/edit#gid=0), the Contig can either be `GenBank` or `RefSeq` accession. So I wanted to bring to attention two things: 
       - I added a couple extra mock lines to `assembly_report.txt` to do complete testing. However, I wasn't sure if this is right, because the pipeline is for Human only. So, only cases for Human assembly report should be tested? 
        - Please review if complete assembly report is required. (It would be handy to have as a default report to work with)
- Add resources
- Little refactoring of resources directory

#### Testing
- Add unit tests. Sample variant object *AFTER* processing:
Contig: CM000686.2   ([source](https://gist.github.com/108krohan/edb9360826a9ee8321063f020d01b533))

- For reference, sample variant object *BEFORE* processing:
Contig: NC_000024.10 ([source](https://gist.github.com/108krohan/a6696a77c07c5a1897c493f096f17cba))


sundarvenkata-EBI:
>I added a couple extra mock lines to assembly_report.txt to do complete testing. However, I wasn't sure if this is right, because the pipeline is for Human only. So, only cases for Human assembly report should be tested?

Let's keep in all the test-cases. Since **seq_id** is just a string I am not sure if NCBI would be able to enforce any rules on the content of the string. So having the test-cases for all the different types of seq_id should be fine. Might help unearth issues sometime down the line (you never know!!)...

eva-accession-import-dbsnp2/src/main/resources/application.properties
```diff
@@ -1,9 +1,10 @@
 spring.batch.job.names=IMPORT_DBSNP_JSON_VARIANTS_JOB
 
 # DbSNP JSON input source
-parameters.input=src/test/resources/input-files/test-dbsnp.json.bz2
+parameters.input=src/test/resources/input-files/json/test-dbsnp.json.bz2
```
cyenyxe: Properties files in `src/main/resources` should not be populated with test values.

eva-accession-import-dbsnp2/src/test/java/uk/ac/ebi/eva/accession/dbsnp2/io/ContigToGenbankReplacerProcessorTest.java
```diff
@@ -0,0 +1,109 @@
...
+    public void setUp() throws Exception {
+        String fileString = ContigToGenbankReplacerProcessorTest.class.getResource(
+                "/input-files/assembly-report/GCF_000001405.38_GRCh38.p12_assembly_report.txt").toString();
+        ContigMapping contigMapping = new ContigMapping(fileString);
+
+        processor = new ContigToGenbankReplacerProcessor(contigMapping);
+    }
+
+    @Test
+    public void ContigGenbank() throws Exception {
```
cyenyxe: Method names shouldn't start with a capital letter.

eva-accession-import-dbsnp2/src/test/resources/input-files/assembly-report/GCF_000001405.38_GRCh38.p12_assembly_report.txt
```diff
@@ -0,0 +1,640 @@
+# Assembly name:  GRCh38.p12
```
cyenyxe: If the vast majority of the lines in the assembly report are not used, could they be cut off of the file?

eva-accession-import-dbsnp2/src/test/resources/application.properties
```diff
@@ -1,9 +1,10 @@
 spring.batch.job.names=IMPORT_DBSNP_JSON_VARIANTS_JOB
 
 # dbSNP JSON input source
-parameters.input=src/test/resources/input-files/test-dbsnp.json.bz2
+parameters.input=src/test/resources/input-files/json/test-dbsnp.json.bz2
```
cyenyxe: How many input files do we have? Because if it's just the 2 (one JSON and one assembly report), I think it may not be worth creating the subfolders.

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/io/ContigToGenbankReplacerProcessor.java
```diff
@@ -0,0 +1,75 @@
...
+import uk.ac.ebi.eva.accession.core.persistence.DbsnpClusteredVariantEntity;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Converts the contig to its GenBank synonym when possible. If the synonym can't be determined it keeps the contig as
+ * is
+ */
+public class ContigToGenbankReplacerProcessor
```
cyenyxe: @jmmut @sundarvenkata-EBI should [ContigReplacerProcessor](https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/ContigReplacerProcessor.java) (from import module) be used as reference instead of [ContigToGenbankReplacerProcessor](https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-pipeline/src/main/java/uk/ac/ebi/eva/accession/pipeline/steps/processors/ContigToGenbankReplacerProcessor.java) (pipeline module)?

sundarvenkata-EBI: The one from the import module deals with SubsnpNoHGVS and its attributes as opposed to IVariant (which is what we have for this pipeline) which the pipeline module uses .

eva-accession-import-dbsnp2/src/test/resources/input-files/assembly-report/GCF_000001405.38_GRCh38.p12_assembly_report.txt
108krohan: Because the file isn't too huge, I favour to bundle this one with the code. Otherwise, the unused lines can be cropped

eva-accession-import-dbsnp2/src/main/java/uk/ac/ebi/eva/accession/dbsnp2/io/ContigToGenbankReplacerProcessor.java
cyenyxe: I'm talking more about the business logic that deals with the assembly report, which is considerably more complex in `ContigReplacerProcessor`, but that is necessary to get dbSNP 1.0 data correctly imported.

sundarvenkata-EBI: @108krohan Could you check if you can adapt the logic from https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/ContigReplacerProcessor.java for this processor? the attributes may differ since that processor uses SubSNPNoHGVS object but could you check if you can supply the analogous attributes from IVariant?

108krohan: As per my assessment, there doesn't seem to be any added benefit of using `ContigToGenbankReplacerProcessor` - as `DbsnpClusteredVariantEntity` object being processed does not have any `Chromosome` field - it has a `Contig` attribute only. Because this is so, the code that checks `Chromosome` attribute is not required. The business logic of `ContigToGenbankReplacerProcessor` dealing with `Contig` replacement with `Genbank` accession is near identical to the one being used in `ContigReplacerProcessor`.

So I would suggest that `ContigReplacerProcessor` be used as reference.

sundarvenkata-EBI: I took a look and looks to me like [isContigReplaceable](https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-import/src/main/java/uk/ac/ebi/eva/accession/dbsnp/processors/ContigSynonymValidationProcessor.java#L87) is pretty much equivalent to [isGenbankReplacementPossible](https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/contig/ContigMapping.java#L163). 

My only opinion is that ContigNotFound should lead to Exception as opposed to just a warning.

jmmut: sorry for the delay answering this. In general, `ContigMapping.isGenbankReplacementPossible`  (which is used in the accession pipeline) should be used to replace to genbank. I confirm that ContigReplacerProcessor mainly focuses on handling both fields contig and chromosome, but also is more permissive on what replacements can be done in a corner case.

Looking at the code it seems you started copying the structure of ContigToGenbankReplacerProcessor and change it to process the approppiate model, using `ContigMapping.isGenbankReplacementPossible`, which is exactly what I recommend doing.

As sundar says, I would also throw an exception if a replacement is not possible. This allows to detect any unmet assumption we might have about human contigs.

PR #159, opened 2019-08-07
EVA-1598 Filter variants for VCF release file of current variants

andresfsilva:
Given that the VCF release file for **current variants** include the alleles and those are only present in the submitted variants a join between the two collections is needed but given that the same RS ID can be mapped against multiple locations (chromosome and/or start position) we will need to filter the results with only the ones that match the location.

There will be an exceptional case with the ambiguous INDELs because dbSNP and the EVA treat them in different ways: dbSNP remove the context nucleotide before an INDEL while the EVA removes the rightmost bases

During the dbSNP import we re-normalize the submitted variants (which can imply modifying the start position) but the clustered variants are not renormalized.

This situation will be handled in this release pipeline by considering the start position as the "same" if the clustered and submitted variants positions:
- Exactly match
- Have a difference of only 1

### DIFF
--- a/eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/AccessionedVariantMongoReader.java
+++ b/eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/AccessionedVariantMongoReader.java
@@ -26,6 +26,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.batch.item.ExecutionContext;
 import org.springframework.batch.item.ItemStreamException;
 
+import uk.ac.ebi.eva.accession.core.SubmittedVariant;
 import uk.ac.ebi.eva.commons.core.models.VariantType;
 import uk.ac.ebi.eva.commons.core.models.VariantTypeToSOAccessionMap;
 import uk.ac.ebi.eva.commons.core.models.pipeline.Variant;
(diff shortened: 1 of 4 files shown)


eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/AccessionedVariantMongoReader.java
```diff
@@ -95,8 +101,36 @@ public void open(ExecutionContext executionContext) throws ItemStreamException {
...
+     * the difference should be 1
+     *
+     * Example:
+     * RS (assembly: GCA_000309985.1, accession: 268233057, chromosome: CM001642.1, start: 7356605, type: INS)
+     * SS (assembly: GCA_000309985.1, accession: 490570267, chromosome: CM001642.1, start: 7356604, reference: ,
+     *     alternate: AGAGCTATGATCTTCGGAAGGAGAAGGAGAAGGAAAAGATTCATGACGTCCAC)
+     */
+    private boolean isSameStart(long clusteredVariantStart, long submittedVariantStart) {
+         return clusteredVariantStart == submittedVariantStart
+                 || Math.abs(clusteredVariantStart - submittedVariantStart) == 1L;
```
sundarvenkata-EBI: Just to be extra careful, can we also add a check for the variant type (INS or DEL) along with the difference check?

andresfsilva: Yes, I think that's a good idea. Check added

PR #160, opened 2019-08-13
EVA-1633 Version control of script to run the release pipeline

jmmut:
When we want to release an assembly, we can call this script to generate the properties file and the jobs that need to be run.


eva-accession-release/src/main/resources/create_release_properties.sh
```diff
@@ -0,0 +1,104 @@
...
+	OUTPUT_FILE=application.properties
+	if [ -f ${OUTPUT_FILE} ]; then
+		mv ${OUTPUT_FILE} ${OUTPUT_FILE}.bk
+		echo "# Made a backup of \"${OUTPUT_FILE}\" into \"${OUTPUT_FILE}.bk\"";
+		if [ ${OUTPUT_FILE} = ${IMPORT_FILE} ]; then
+			IMPORT_FILE=${IMPORT_FILE}.bk
+		fi
+	fi
+
+	# extract values from the existing properties file
```
sundarvenkata-EBI: Except for assemblyAccession values for other variables don't seem to be used elsewhere except for the new properties file. So why not copy the lines verbatim for others instead of using "sed"?

eva-accession-release/src/main/resources/create_release_properties.sh
```diff
@@ -0,0 +1,104 @@
...
+	assemblyAccession=`grep -m 1 "^parameters.assemblyAccession" $IMPORT_FILE | sed "s/^[^=]\+=//"`
```
sundarvenkata-EBI: Does this strip out whitespaces too?

eva-accession-release/src/main/resources/create_release_properties.sh
```diff
@@ -0,0 +1,104 @@
...
+jobname=\`date +\"%Y%m%d_%H%M%S\"\`_${assemblyAccession}
+bsub -J \${jobname} -o release.log -e release.err -M 8192 -R \"rusage[mem=8192]\" \"java -Xmx7g -jar ${EVA_ACCESSION_JAR_PATH}\"
+bsub -J \${jobname}_sort -w \${jobname} -o sort_ids.log -e sort_ids.err -M 8192 -R \"rusage[mem=8192]\" \"sort -V ${OUTPUT_FOLDER}${assemblyAccession}_deprecated_ids.unsorted.txt | uniq | gzip > ${OUTPUT_FOLDER}${assemblyAccession}_deprecated_ids.txt.gz; sort -V ${OUTPUT_FOLDER}${assemblyAccession}_merged_deprecated_ids.unsorted.txt | uniq | gzip > ${OUTPUT_FOLDER}${assemblyAccession}_merged_deprecated_ids.txt.gz\"
+bsub -J \${jobname}_validate -w \${jobname} -o ${OUTPUT_FOLDER}validate.log -e ${OUTPUT_FOLDER}validate.err \"${VCF_VALIDATOR_PATH} -i ${OUTPUT_FOLDER}${assemblyAccession}_current_ids.vcf -o ${OUTPUT_FOLDER}; ${VCF_VALIDATOR_PATH} -i ${OUTPUT_FOLDER}${assemblyAccession}_merged_ids.vcf -o ${OUTPUT_FOLDER}; ${VCF_ASM_CHECKER_PATH} -i ${OUTPUT_FOLDER}${assemblyAccession}_current_ids.vcf -f ${fasta} -o ${OUTPUT_FOLDER} -r text,summary ; ${VCF_ASM_CHECKER_PATH} -i ${OUTPUT_FOLDER}${assemblyAccession}_merged_ids.vcf -f ${fasta} -o ${OUTPUT_FOLDER} -r text,summary ;\"
```
sundarvenkata-EBI: We are not waiting for the sort to finish. So, if the validation fails and we need to regenerate the VCF for some reason, won't the sort go to waste (may be this is rare and we don't need to worry about this)? Also, do we get a lot of value running these processes in parallel?  If there are intermediate failures, this might aggravate the number of stale jobs that are spawned and might make tracking difficult...

jmmut: Actually, there's another problem. I only made the compression depend on the validation so that it waits to delete the original VCF. We will have to compress and release some VCFs that won't pass the validator, like having duplicates.
But now I realise that if the validation job fails, the others won't start. I have to change this so that they start if the validation job is finished or failed.

PR #161, opened 2019-08-13
EVA-1639 Check ss ids declustered operations before throwing exeption

andresfsilva:
An exception should not be thrown if a clustered variant was merged but there are not operation for the submitted variants because they were declustered before. 

This happens because the decluster operations are created before the merge operations.

So we will have some situations where there is a merge operation but the merged clustered variant does not have any submitted variants associated because they all have been declustered (rs = null). Hence no update operation to indicate the merge is created for submitted variants.

### DIFF
--- a/eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/MergedVariantMongoReader.java
+++ b/eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/MergedVariantMongoReader.java
@@ -106,7 +106,7 @@ public class MergedVariantMongoReader extends VariantMongoAggregationReader {
         String sequenceOntology = VariantTypeToSOAccessionMap.getSequenceOntologyAccession(VariantType.valueOf(type));
         boolean validated = inactiveEntity.getBoolean(VALIDATED_FIELD);
 
-        Map<String, Variant> variants = new HashMap<>();
+        Map<String, Variant> mergedVariants = new HashMap<>();
         Collection<Document> submittedVariantOperations = (Collection<Document>) mergedVariant.get(SS_INFO_FIELD);
 
         boolean hasSubmittedVariantsDeclustered = false;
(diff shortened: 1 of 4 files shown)


eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/MergedVariantMongoReader.java
```diff
@@ -135,11 +138,19 @@ private String getInactiveField(String field) {
...
+            }
         }
 
-        if (variants.isEmpty()) {
-            throw new IllegalStateException("There was a merge operation for rs" + rs + " but there is no update " +
-                                                    "operation for the SS IDs.");
+        if (!hasSubmittedVariantsDeclustered && variants.isEmpty()) {
+            throw new IllegalStateException ("There was a merge operation for rs" + rs + " but there is no update " +
+                                                     "operation for the SS IDs. This can be either an error or a " +
+                                                     "special scenario where all the corresponding SS IDs were " +
```
sundarvenkata-EBI: Isn't hasSubmittedVariantsDeclustered supposed to indicate precisely this condition?

andresfsilva: You are right! Message have been updated

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/MergedVariantMongoReader.java
```diff
@@ -135,11 +138,21 @@ private String getInactiveField(String field) {
...
+                    && submittedVariantOperation.getString(REASON).startsWith(DECLUSTER_OPERATION_REASON_FIRST_WORD)) {
+                hasSubmittedVariantsDeclustered = true;
+            }
         }
 
-        if (variants.isEmpty()) {
-            throw new IllegalStateException("There was a merge operation for rs" + rs + " but there is no update " +
-                                                    "operation for the SS IDs.");
+        if (!hasSubmittedVariantsDeclustered && variants.isEmpty()) {
+            throw new IllegalStateException ("There was a merge operation for rs" + rs + " but there is no update " +
```
cyenyxe: Suggested simplification of this message:
> "Found merge operation for rs" + rs + " but no SS IDs updates (merge/update) in the collection containing operations. Since every RS ID in dbSNP is associated with at least one SS ID, the latter must be updated when the former are merged, unless they have been previously declustered (RS ID = null)."

PR #166, opened 2019-08-22
EVA-1644 Use assembly names instead of asssembly accessions

jmmut:
Pending the manual check of community databases outside of ENA. Maybe the assembly names that ENA provides are not the most common one used in those communities, but for the case where ENA names are valid, the code is complete.

Edit: only grape (GCA_000003745.2) was found that the same genome is present in ENA and the databases with different assembly names. The name of the databases will be prioritized.


eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/AssemblyNameRetriever.java
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * However, the alias is optional, according to
+ * https://github.com/enasequence/schema/blob/master/src/main/resources/uk/ac/ebi/ena/sra/schema/ENA.assembly.xsd#L32
+ * which defines an AssemblyType as an extension of ObjectType, and
+ * https://github.com/enasequence/schema/blob/master/src/main/resources/uk/ac/ebi/ena/sra/schema/SRA.common.xsd#L21
+ * defines alias as optional attribute of ObjectType.
+ */
+public class AssemblyNameRetriever {
+
+    private static final String ENA_ASSMEBLY_URL_FORMAT_STRING = "https://www.ebi.ac.uk/ena/data/view/%s";
```
andresfsilva: Typo in `ASSEMBLY` for this and the next constant

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/AssemblyNameRetriever.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        public Assembly getAssembly() {
+            return assembly;
+        }
+
+        static class Assembly {
+
+            @XmlAttribute
+            private String alias;
```
andresfsilva: There is a `NAME` attribute in the XML response. Is it better to use the alias instead of the name?
```
<ASSEMBLY accession="GCA_000004515.3" alias="Glycine_max_v2.0" center_name="US DOE Joint Genome Institute (JGI-PGF)">
[...]
    <NAME>Glycine_max_v2.0</NAME>
```

sundarvenkata-EBI: Ensure that NAME attribute is mandatory as part of the XSD and you can also check if it has expected values for a few species.

jmmut: I checked that it's not optional and it gives the same results as the alias for all our species, so I'll use the NAME

eva-accession-release/src/test/java/uk/ac/ebi/eva/accession/release/io/AssemblyNameRetrieverTest.java
```diff
@@ -0,0 +1,89 @@
...
+        fileWriter.close();
+
+        JAXBContext jaxbContext = JAXBContext.newInstance(EnaAssemblyXml.class);
+        Unmarshaller unmarshaller = jaxbContext.createUnmarshaller();
+        EnaAssemblyXml enaAssembly = (EnaAssemblyXml) unmarshaller.unmarshal(xml);
+        assertEquals("GRCh38.p13", enaAssembly.getAssembly().getName());
+    }
+
+    @Test
+    public void parseMissingAlias() throws IOException, JAXBException {
```
andresfsilva:
```suggestion
    public void parseMissingName() throws IOException, JAXBException {
```

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/AssemblyNameRetriever.java
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * Use a curated list and the ENA webservices to return the assembly name associated with an assembly accession.
+ *
+ * The ENA webservices can return an xml that starts like this:
+ * <pre>
+ * {@code
+ * <?xml version="1.0" encoding="UTF-8"?>
+ * <ROOT request="GCA_000001405.28&amp;display=xml">
+ * <ASSEMBLY accession="GCA_000001405.28" alias="GRCh38.p13" center_name="Genome Reference Consortium">
```
andresfsilva: Can you also include the name tag given that we are using the name instead of the alias?
Also not sure if the below description should changed to not talk about the alias but the name

PR #167, opened 2019-08-27
EVA-1651 Handle variant start greater than chromosome end

andresfsilva:
When the start position of a variant is greater than the length of the chromosome the pipeline should skip such variant and log the message indicating this happened.

The listener will log the total counts in the `afterStep` method as follows:
- Invalid variants (filtered out in `ExcludeInvalidVariantsProcessor`) as **filtered out** 
- Variants with start greater than chromosome end as **skipped**


eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/listeners/ExcludeVariantsListener.java
```diff
@@ -31,6 +31,8 @@
     @Override
     public ExitStatus afterStep(StepExecution stepExecution) {
         logger.info("Processors filtered out {} variants", stepExecution.getFilterCount());
+        logger.warn("Processors skipped {} variants because the start position is greater than the chromosome end",
+                    stepExecution.getSkipCount());
```
jmmut: if we are not writing the list of contigs affected here, we don't need the variantsWithIllegalStart set in the processor.

PR #168, opened 2019-09-03
EVA-1656 Modify the script to include sort and the assembly report for validator

andresfsilva:
Modified the `create_release_properties.sh` script to:

- Include the assembly report as part of the assembly checker command (-a parameter)
- Sort VCF files before running the validators. This because files can get unsorted when adding the context nucleotides required by VCF format.
- Modified jobs for current, merged, and txt files to be run in parallel

Release pipeline: generates current, merged and txt files

Txt files: Sort
Current VCF: Sort - Validate - Compress
Merged VCF: Sort - Validate - Compress

Tabix: Dependant on the VCF jobs
Counts: Dependant on the VCF and txt jobs


eva-accession-release/src/main/resources/create_release_properties.sh
```diff
@@ -94,12 +96,12 @@ logging.level.uk.ac.ebi.eva.accession.release=INFO
...
+bsub -J \${jobname}_tabix -w \"(\${jobname}_current_vcf && \${jobname}_merged_vcf)\" -o tabix.log -e tabix.err \"${TABIX_PATH} ${OUTPUT_FOLDER}${assemblyAccession}_current_ids.vcf.gz; ${TABIX_PATH}  ${OUTPUT_FOLDER}${assemblyAccession}_merged_ids.vcf.gz\"
+bsub -J \${jobname}_count -w \"(\${jobname}_txts && \${jobname}_current_vcf && \${jobname}_merged_vcf)\" -o count_ids.log -e count_ids.err \"echo '# Unique RS ID counts' > ${OUTPUT_FOLDER}README_rs_ids_counts.txt; ${COUNT_IDS_PATH} ${OUTPUT_FOLDER}${assemblyAccession}_current_ids.vcf.gz >> ${OUTPUT_FOLDER}README_rs_ids_counts.txt; ${COUNT_IDS_PATH} ${OUTPUT_FOLDER}${assemblyAccession}_merged_ids.vcf.gz >> ${OUTPUT_FOLDER}README_rs_ids_counts.txt; zcat ${OUTPUT_FOLDER}${assemblyAccession}_deprecated_ids.txt.gz | wc -l | sed 's/^/'${assemblyAccession}_deprecated_ids.txt.gz'\t/' >> ${OUTPUT_FOLDER}README_rs_ids_counts.txt; zcat ${OUTPUT_FOLDER}${assemblyAccession}_merged_deprecated_ids.txt.gz | cut -f 1 | uniq | wc -l | sed 's/^/'${assemblyAccession}_merged_deprecated_ids.txt.gz'\t/' >> ${OUTPUT_FOLDER}README_rs_ids_counts.txt\"
```
jmmut: I know you didn't change most of this line, but why are we not sorting the deprecated_ids and merged_deprecated_ids files? I cannot see any sort in the release pipeline.

andresfsilva: Those `.txt` files are sorted [here](https://github.com/EBIvariation/eva-accession/pull/168/files#diff-a8b61f207cb67bd5eb10ad0a5b586b2cR100)

eva-accession-release/src/main/resources/create_release_properties.sh
```diff
@@ -94,12 +96,12 @@ logging.level.uk.ac.ebi.eva.accession.release=INFO
...
+bsub -J \${jobname}_txts -w \${jobname} -o txts.log -e txts.err -M 8192 -R \"rusage[mem=8192]\" \"sort -V ${OUTPUT_FOLDER}${assemblyAccession}_deprecated_ids.unsorted.txt | uniq | gzip > ${OUTPUT_FOLDER}${assemblyAccession}_deprecated_ids.txt.gz; sort -V ${OUTPUT_FOLDER}${assemblyAccession}_merged_deprecated_ids.unsorted.txt | uniq | gzip > ${OUTPUT_FOLDER}${assemblyAccession}_merged_deprecated_ids.txt.gz\"
+bsub -J \${jobname}_current_vcf -w \${jobname} -o ${OUTPUT_FOLDER}current_vcf.log -e ${OUTPUT_FOLDER}current_vcf.err \"${SORT_SCRIPT_PATH} ${OUTPUT_FOLDER}${assemblyAccession}_current_ids_unsorted.vcf ${OUTPUT_FOLDER}${assemblyAccession}_current_ids.vcf; ${VCF_VALIDATOR_PATH} -i ${OUTPUT_FOLDER}${assemblyAccession}_current_ids.vcf -o ${OUTPUT_FOLDER}; ${VCF_ASM_CHECKER_PATH} -i ${OUTPUT_FOLDER}${assemblyAccession}_current_ids.vcf -f ${fasta} -a ${assembly_report} -o ${OUTPUT_FOLDER} -r text,summary ; ${BGZIP_PATH} ${OUTPUT_FOLDER}${assemblyAccession}_current_ids.vcf\"
```
sundarvenkata-EBI: Looks like we have combined validate and compress into a single command. Did you check if compress runs if the validation fails?

sundarvenkata-EBI: Also, have you checked if the sort produces chromosomes in the order that Thomas suggested (lexicographically)?

jmmut: the commands are split by `;` so they should run regardless of the exit code of the previous command. And this is intended because we want to compress even if the validator fails, because we know there are some non-VCF-compliant issues

andresfsilva: We run the `VCF validation` and the `assembly checker` but do not actually read the file automatically to see if everything went well, we check it manually afterwards but we don't stop the process. For instance the VCF validator will have some messages about duplicate variants, which is supposed to be an error in VCFs, but that will not be an error in our case. So yes, the process will continue

andresfsilva: > Also, have you checked if the sort produces chromosomes in the order that Thomas suggested (lexicographically)?

Yes, the `sort_vcf` script was modified to write the chromosomes in lexicographical order.

PR #172, opened 2019-09-13
EVA-1665 VCF headers list the contigs in accession order

jmmut:
Of course this is platform dependent, but given that we have some
species with lots of contigs (e.g. almost 0.5M for rat or turkey),
we can either:

- keep the hacky implementation of this commit.
- evaluate memory performance of: loading the file into
a list, sort the list in memory, write the list. For at least a
couple million of 20 bytes strings.

If we go for the second option, the changes have to be done in the
same place as the current system sort.

### DIFF
--- a/eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/ContigWriter.java
+++ b/eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/ContigWriter.java
@@ -69,6 +69,19 @@ public class ContigWriter implements ItemStreamWriter<String> {
     @Override
     public void close() throws ItemStreamException {
         printWriter.close();
+        sortContigFile();
+    }
+
+    private void sortContigFile() {
+        try {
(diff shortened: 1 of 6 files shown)


eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/io/ContigWriter.java
```diff
@@ -69,6 +71,30 @@ public void update(ExecutionContext executionContext) throws ItemStreamException
     @Override
     public void close() throws ItemStreamException {
         printWriter.close();
+        sortContigFile();
+    }
+
+    private void sortContigFile() {
+        try {
+            Process process = Runtime.getRuntime().exec(new String[]{"sort", this.output.getAbsolutePath()});
```
jmmut: instead of reading sort's standard output, why not using 
```suggestion
            String[] command = {"sort", "-o", this.output.getAbsolutePath(), this.output.getAbsolutePath()};
            Process process = Runtime.getRuntime().exec(command);
```
?
it was my bad to use `sort` without arguments but this is what I had in mind from the beginning.

PR #176, opened 2019-10-01
EVA-1666 Release the species other than EVA-1651

jmmut:
parsing asm report, overwrite duplicate genbank only if it's assembled-molecule.

What this achieves is ignoring scaffolds when translating from
genbank to other name (e.g. from genbank to sequence name), while
still having the mappings from different names to the same genbank.


jmmut:
andres confirmed the code works as expected with a species released in the cluster

PR #184, opened 2019-12-12
EVA-1760 Support human variants query by id fields

andresfsilva:


eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/service/DbsnpClusteredHumanVariantAccessioningService.java
```diff
@@ -61,4 +67,27 @@ public DbsnpClusteredHumanVariantAccessioningService(@Qualifier("humanActiveServ
...
+        IClusteredVariant clusteredVariantToSearch = new ClusteredVariant(assembly, 0, contig, start, type, false,
+                                                                          null);
+
+        List<AccessionWrapper<IClusteredVariant, String, Long>> activeClusteredVariantWrappers =
+                humanService.get(Collections.singletonList(clusteredVariantToSearch));
+        List<AccessionResponseDTO<ClusteredVariant, IClusteredVariant, String, Long>> activeClusteredVariants =
+                activeClusteredVariantWrappers.stream()
+                                              .map(v -> new AccessionResponseDTO<>(v, ClusteredVariant::new))
+                                              .collect(Collectors.toList());
+        clusteredVariants.addAll(activeClusteredVariants);
```
sundarvenkata-EBI: Intellij suggests ``List<AccessionResponseDTO<ClusteredVariant, IClusteredVariant, String, Long>> clusteredVariants =
                new ArrayList(activeClusteredVariants);`` and I think that is a good idea too...

eva-accession-ws/src/test/java/uk/ac/ebi/eva/accession/ws/ClusteredVariantsRestControllerTest.java
```diff
@@ -811,19 +811,45 @@ public void testGetDeprecatedEvaClusteredVariant()
...
+    public void findByIdFieldsHumanVariantInOperations() {
+        ResponseEntity<List<AccessionResponseDTO<ClusteredVariant, IClusteredVariant, String, Long>>> getVariantsResponse =
+                controller.getByIdFields(clusteredHumanVariantEntity3.getAssemblyAccession(),
+                                         clusteredHumanVariantEntity3.getContig(),
+                                         clusteredHumanVariantEntity3.getStart(),
+                                         clusteredHumanVariantEntity3.getType());
+
+        assertEquals(HttpStatus.OK, getVariantsResponse.getStatusCode());
+        assertEquals(clusteredHumanVariantEntity3.getAccession(), getVariantsResponse.getBody().get(0).getAccession());
+    }
```
sundarvenkata-EBI: Should we have another test to check responses for non-existent human variants?

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/service/DbsnpClusteredHumanVariantAccessioningService.java
andresfsilva: I kind of like better using the `addAll` because we are adding `activeClusteredVariants` and `activeClusteredVariantsInOperations` to the `clusteredVariants` List, so for me it's easier to read if we use the same method to add in both cases. What do you think about it @sundarvenkata-EBI ?

sundarvenkata-EBI: Sure. It was just a judgement call... I am fine with leaving this as it stands...

PR #186, opened 2020-01-28
EVA-1745 Changed the package structure in eva-accession-core module

andresfsilva:


eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/configuration/human/MongoHumanConfiguration.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
```
sundarvenkata-EBI: MongoHumanConfiguration seems to be redundant since the configuration file is under the human folder anyway. Are there technical constraints in renaming this configuration file to  MongoConfiguration?

andresfsilva: There are not technical constraints, but having the classes with the same name, I think, could become confusing when reading the code. The way I see it, it will be `non human` unless the class name says `human`. The same way we did for `dbsnp` and `eva` classes.

I agree with you that using `Human` at the beginning of the class names can be clearer (We did that for dbsnp classes). There is a step in the ticket for renaming classes after the package structure refactoring is finished for all modules.

PR #189, opened 2020-02-05
EVA-1745 Return models in core module

andresfsilva:
Return MODELS in the `core` module and convert them to DTOs in the `ws` module

### DIFF
--- a/eva-accession-ws/src/main/java/uk/ac/ebi/eva/accession/ws/rest/ClusteredVariantsRestController.java
+++ b/eva-accession-ws/src/main/java/uk/ac/ebi/eva/accession/ws/rest/ClusteredVariantsRestController.java
@@ -193,10 +193,9 @@ public class ClusteredVariantsRestController {
                 nonHumanActiveService.getByIdFields(assembly, chromosome, start, variantType);
         clusteredVariantWrapper.map(this::toDTO).ifPresent(clusteredVariants::add);
 
-        Optional<List<AccessionWrapper<IClusteredVariant, String, Long>>> humanClusteredVariants =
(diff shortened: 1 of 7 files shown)


eva-accession-ws/src/main/java/uk/ac/ebi/eva/accession/ws/rest/ClusteredVariantsRestController.java
```diff
@@ -187,17 +189,23 @@ public ClusteredVariantsRestController(
...
+        Optional<List<AccessionWrapper<IClusteredVariant, String, Long>>> humanClusteredVariants =
                 humanService.getByIdFields(assembly, chromosome, start, variantType);
-        humanClusteredVariant.ifPresent(clusteredVariants::addAll);
+        humanClusteredVariants.ifPresent(wrappers -> clusteredVariants.addAll(
+                wrappers.stream().map(this::toDto).collect(Collectors.toList())));
 
         return clusteredVariants.isEmpty() ? ResponseEntity.notFound().build() : ResponseEntity.ok(clusteredVariants);
     }
 
+    private AccessionResponseDTO<ClusteredVariant, IClusteredVariant, String, Long> toDto(
```
sundarvenkata-EBI: Minor quibble here... Is toDTO acceptable when it comes to camel case? that seems more readable... I almost read this as todo...

jmmut: I would vote for `toDto` for consistency in acronyms across EVA java code, but given that in accession commons they used `AccessionResponseDTO` (used here), I am ok prioritising local consistency over global consistency.

PR #190, opened 2020-02-06
EVA-1818 - Monitor duplicate accessions in MongoDB

sundarvenkata-EBI:
monitor_duplicate_accessions.py - Given a collection name and database properties file, export accessions to a file and identify duplicates.

### DIFF
--- a/eva-accession-pipeline/src/main/resources/monitoring/monitor_duplicate_accessions.py
+++ b/eva-accession-pipeline/src/main/resources/monitoring/monitor_duplicate_accessions.py
@@ -23,7 +23,8 @@ from __init__ import *
 
 def export_mongo_accessions(mongo_connection_properties, collection_name, export_output_filename):
     export_command = 'mongoexport --host {0} --port {1} --db {2} --username {3} --password {4} ' \
-                     '--authenticationDatabase={5} --collection {6} --type=csv --fields accession -o "{7}" '\
+                     '--authenticationDatabase={5} --collection {6} --type=csv --fields accession -o "{7}" ' \
+                     '--sort \'{{accession: 1}}\''\
                     .format(mongo_connection_properties["mongo_host"],
                             mongo_connection_properties["mongo_port"],
                             mongo_connection_properties["mongo_db"],
@@ -31,7 +32,7 @@ def export_mongo_accessions(mongo_connection_properties, collection_name, export
                             mongo_connection_properties["mongo_password"],
                             mongo_connection_properties["mongo_auth_db"],
                             collection_name,
-                            export_output_filename) + " --sort '{accession: 1}'"
+                            export_output_filename)
     run_command_with_output("Exporting accessions in the {0} collection in the {1} database at {2}..."
                             .format(collection_name, mongo_connection_properties["mongo_db"],
                                     mongo_connection_properties["mongo_host"]),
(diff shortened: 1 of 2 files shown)


eva-accession-pipeline/src/main/resources/monitoring/monitor_duplicate_accessions.py
```diff
@@ -0,0 +1,105 @@
...
+    export_command = 'mongoexport --host {0} --port {1} --db {2} --username {3} --password {4} ' \
+                     '--authenticationDatabase={5} --collection {6} --type=csv --fields accession -o "{7}" '\
+                    .format(mongo_connection_properties["mongo_host"],
+                            mongo_connection_properties["mongo_port"],
+                            mongo_connection_properties["mongo_db"],
+                            mongo_connection_properties["mongo_username"],
+                            mongo_connection_properties["mongo_password"],
+                            mongo_connection_properties["mongo_auth_db"],
+                            collection_name,
+                            export_output_filename) + " --sort '{accession: 1}'"
```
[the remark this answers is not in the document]
sundarvenkata-EBI: The previous string is delimited by single quotes and the json string argument for sort is also single quotes so it would have to be escaped... i felt that it might look messy that way so I separated them...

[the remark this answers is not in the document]
jmmut: I pointed it out because I was expecting a sort somewhere and I didn't find it at first. IMO keeping that together is more important than avoiding quote mess.

I would prefer either:
- escaping: 
```
'single \'quotes\'' \
'double "quotes"'
```
- or using triple quotes (note the last space to avoid 4 quotes):
```
"""single \'quotes\'' \
"""double "quotes" """
```
- or changing the outer quotes from each line.
```
"single 'quotes'"
'double "quotes"'
```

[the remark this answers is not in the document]
sundarvenkata-EBI: Resolved...

eva-accession-pipeline/src/main/resources/monitoring/monitor_duplicate_accessions.py
```diff
@@ -0,0 +1,106 @@
...
+from datetime import datetime
+import getpass
+import smtplib
+from __init__ import *
+
+
+def export_mongo_accessions(mongo_connection_properties, collection_name, export_output_filename):
+    export_command = 'mongoexport --host {0} --port {1} --db {2} --username {3} --password {4} ' \
+                     '--authenticationDatabase={5} --collection {6} --type=csv --fields accession -o "{7}" ' \
+                     '--sort \'{{accession: 1}}\''\
```
tcezard: I'm wondering if it's better to sort on the server (putting pressure on the server) or to move the sorting to the output file which would run wherever the script is run.

jmmut: In theory this is a [covered query](https://docs.mongodb.com/manual/core/query-optimization/#covered-query), so it should run just scanning the "accession" index, and the indexes are usually kept in memory, so it shouldn't put much pressure. However, I wouldn't put my hand over the fire about it, as MongoDB is surprisingly picky on this stuff.

sundarvenkata-EBI: the covered query does run without any issues on PROD and it is covered by the index. If not, mongoexport will fail per specification if the output is more than 100MB - see [here](https://docs.mongodb.com/master/reference/program/mongoexport/#cmdoption-mongoexport-sort)

tcezard: ok

PR #194, opened 2020-03-03
EVA-1822 Automate identifiers web services deployment with gitlab

andresfsilva:


.gitlab-ci.yml
```diff
@@ -0,0 +1,120 @@
...
+before_script:
+    - export DATETIME=$(date +%Y-%m-%dT%H-%M-%S)
+
+test:
+    stage: test
+    image: maven:3.6.1-jdk-8-alpine
+    services:
+        - mongo:3.2.17
+    script:
+        - mvn clean test -pl 'eva-accession-core,eva-accession-ws' -Deva.mongo.host.test=mongo
```
sundarvenkata-EBI: Judgement call: The non-Unix convention of maven command line switch is a bit confusing since -pl constitutes one switch. Could we expand to the more intuitive --projects?
```suggestion
        - mvn clean test --projects 'eva-accession-core,eva-accession-ws' -Deva.mongo.host.test=mongo
```

andresfsilva:
I grouped some commits so we don't have that many and the process is clear

PR #196, opened 2020-03-05
EVA-1838 clustering processor prototype

andresfsilva:


eva-accession-cluster/pom.xml
```diff
@@ -0,0 +1,22 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>eva-accession</artifactId>
+        <groupId>uk.ac.ebi.eva</groupId>
+        <version>0.3.2</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>eva-accession-cluster</artifactId>
```
sundarvenkata-EBI: Judgement call: May be we should name the package **eva-accession-clustering** to be more clear?

jmmut: I thought that too, but at this point the word has lost its meaning to me, out of repetition.

eva-accession-cluster/src/main/java/uk/ac/ebi/eva/accession/cluster/batch/processors/ClusterVariantProcessor.java
```diff
@@ -0,0 +1,90 @@
...
+                                                                     submittedVariant.getContig(),
+                                                                     submittedVariant.getStart(),
+                                                                     getVariantType(
+                                                                             submittedVariant.getReferenceAllele(),
+                                                                             submittedVariant.getAlternateAllele()),
+                                                                     submittedVariant.isValidated(),
+                                                                     submittedVariant.getCreatedDate());
+            String hash = hashingFunction.apply(clusteredVariant);
+            if (assignedAccessions.containsKey(hash)) {
+                Long existingClusteredVariantAccession = assignedAccessions.get(hash);
```
jmmut: I have lots of ideas on how to improve this mockup to be closer to a production usage, and also allow easy testing, but for now the prototype is ok, we can iterate from here.

eva-accession-cluster/pom.xml
sundarvenkata-EBI: We will still know what it means... I meant for someone new who is looking at the code :D

andresfsilva: Module renamed to eva-accession-clustering

PR #200, opened 2020-03-16
EVA-1897 Implement accessioning service for clustering pipeline

andresfsilva:


eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/model/dbsnp/DbsnpClusteredVariantEntity.java
```diff
@@ -17,150 +17,30 @@
...
-import uk.ac.ebi.ampt2d.commons.accession.persistence.mongodb.document.AccessionedDocument;
-
-import uk.ac.ebi.eva.accession.core.model.ClusteredVariant;
 import uk.ac.ebi.eva.accession.core.model.IClusteredVariant;
+import uk.ac.ebi.eva.accession.core.model.eva.ClusteredVariantEntity;
 import uk.ac.ebi.eva.commons.core.models.VariantType;
 
 import java.time.LocalDateTime;
 
-@Document
```
jmmut: removed @Document? doesn't need it because the parent has it? have you checked this works?

andresfsilva: Correct, the parent has it so it is not needed. We are doing it the same way with Submitted Variants

PR #201, opened 2020-03-18
EVA-1900 - Clustering deployment

tcezard:


.gitlab-ci.yml
```diff
@@ -59,17 +65,22 @@ package-production:
     only:
         - tags
 
-# Not executed, parent job definition for Tomcat deployments
+# Not executed, parent job definition for deployments
 .deploy:
```
andresfsilva: I think we should separate the `web services` and `clustering pipeline` deployment, that way the deployment of one module will not be affected when the deployment of another module fails.
So we can create `.deploy-ws` and `.deploy-clustering` abstract stages

.gitlab-ci.yml
```diff
@@ -4,7 +4,8 @@ stages:
     - deploy
 
 variables:
-    ARTIFACT_PATH: eva-accession-ws/target/eva-$ENVIRONMENT_NAME.war
+    ACC_WS_ARTIFACT_PATH: eva-accession-ws/target/eva-acc-ws-$ENVIRONMENT_NAME.war
+    ACC_CLUST_ARTIFACT_PATH: eva-accession-clustering/target/eva-acc-clust-$ENVIRONMENT_NAME.jar
```
andresfsilva: I would prefer to keep the variable names with complete words. We could use `WS_ARTIFACT_PATH` and `CLUSTERING_ARTIFACT_PATH` if you don't want them to be that long

.gitlab-ci.yml
```diff
@@ -59,17 +65,22 @@ package-production:
...
     image: alpine:3.9.5
     environment:
         name: $ENVIRONMENT_NAME
     script:
-        - apk add --update curl
-        - curl -u $TOMCAT_USER:$TOMCAT_PASSWORD -T "$ARTIFACT_PATH" "http://$TOMCAT_HOST/manager/text/deploy?update=true&path=/eva/webservices/identifiers&version=$DATETIME" | grep "OK - Deployed application"
+        - apk add --update curl openssh-client
+        - curl -u $TOMCAT_USER:$TOMCAT_PASSWORD -T "$ACC_WS_ARTIFACT_PATH" "http://$TOMCAT_HOST/manager/text/deploy?update=true&path=/eva/webservices/identifiers&version=$DATETIME" | grep "OK - Deployed application"
+        # Start ssh-agent and set identity then copy the Jar file and create a symlink
+        - eval "$(ssh-agent)"; echo "$EVA_ETL_PRIVATE_KEY" > identity_id_rsa; chmod 600 identity_id_rsa; ssh-add identity_id_rsa
```
andresfsilva: Can this be simplified? also it can be split in different lines so its more readable, maybe something like:
```
        - eval $(ssh-agent)
        - echo "$EVA_ETL_PRIVATE_KEY" | ssh-add -
```

.gitlab-ci.yml
```diff
@@ -59,17 +65,22 @@ package-production:
...
+        - export ACC_CLUST_ARTIFACT_NAME=eva-accession-clustering-${ENVIRONMENT_NAME}-${DATETIME}.jar
```
andresfsilva: I think you don't need to use `export` here, with a local variable is enough

.gitlab-ci.yml
```diff
@@ -59,17 +65,22 @@ package-production:
...
+        - if [[ ! -z "$CLUSTERING_PATH" ]]; then scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null $ACC_CLUST_ARTIFACT_PATH ${CLUSTER_USER}@${CLUSTER_LOGIN_NODE}:${CLUSTERING_PATH}/${ACC_CLUST_ARTIFACT_NAME} ; fi
+        - if [[ ! -z "$CLUSTERING_PATH" ]]; then ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null ${CLUSTER_USER}@${CLUSTER_LOGIN_NODE} "ln -fns ${CLUSTERING_PATH}/${ACC_CLUST_ARTIFACT_NAME} ${CLUSTERING_PATH}/eva-accession-clustering.jar"; fi
```
andresfsilva: Do you think its necessary to have the `if` statement? I think we can count on the path being valid at this stage

sundarvenkata-EBI: I think checking the path is warranted in a corner case: NFS screws up (happens sometimes) between the scp and ssh and an invalid symlink is left behind....

sundarvenkata-EBI: Ok, I just checked and **ln** still acts sensibly with -f switch and checks for non-existent files....

.gitlab-ci.yml
tcezard: That's a good point. I  split the deploy stages

.gitlab-ci.yml
tcezard: Done. Although one could argue that `WS_ARTIFACT_PATH` is a contraction as well.

.gitlab-ci.yml
tcezard: That does improve readability

.gitlab-ci.yml
tcezard: The purpose of the if statement was simply to avoid deploying in `deploy-internal` and `deploy-production-fallback`. Now that the WS and jar have separate deployments the statements are not useful and I've removed them.

.gitlab-ci.yml
tcezard: Removed

.gitlab-ci.yml
```diff
@@ -118,3 +122,42 @@ deploy-tomcat-production-fallback:
...
+deploy-jar-development:
+    extends: .deploy-jar
+    variables:
+        ENVIRONMENT_NAME: development
+        CLUSTERING_PATH: $CLUSTERING_DEVELOPMENT_PATH
+    dependencies:
+        - package-development
+    only:
+        - master
+        - clustering_deployment
```
andresfsilva: This should be removed before merging into master

.gitlab-ci.yml
```diff
@@ -20,6 +21,7 @@ test:
         name: test-env
     only:
         - master
+        - clustering_deployment
```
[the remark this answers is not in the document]
tcezard: removed

.gitlab-ci.yml
tcezard: removed

PR #209, opened 2020-04-22
EVA-1915 Fix settings internal

andresfsilva:


andresfsilva:
Not needed anymore, after PR #210 maven properties are downloaded from the configuration private repository and the `.gitlab.settings.xml` file is removed

PR #216, opened 2020-06-21
Skip empty lines when concatenating the fasta file

tcezard:


sundarvenkata-EBI:
Suggest re-basing the PR...

PR #217, opened 2020-06-22
EVA-2003 Make use of the RSid provided in the input VCF

jmmut:
This is a draft PR with all the basic scenarios, which I'm posting for early reviews, but I still want to test more corner cases and polish everything a bit more.

## Required fixes before using in production

- [x] If an RSid already exist in the input VCF we should not assign this existing RS
- [x] Check if there is an existing RSid for this variant location and type.
  - [x] automatic if we use get or create
  - [x] what about EVA SS that cluster into dbsnp RS? the EVA ss will get rs=x where x is not in EVA RS table
    - [x] solved if we finally create the ClusteredVariantService that uses EVA and dbsnp
- [x] Detect the fact that the new ClusteredVariant will have multiple possible RSid
  - [x] don't merge the RS IDs if any of them matches to several positions in the same assembly
- [x] Select one RSid and deprecate the other ones
  - [x] current dbsnp RS collision with DB dbsnp RS
  - [x] current eva RS collision with DB dbsnp RS (or viceversa)
  - [x] current eva RS collision with DB EVA RS
- [x] remove Service::mergeKeepingentries (I added it for this but ended up with a different solution)
- [x] are we going to cluster dbsnp variants?
  - yes, some dbsnp variants have rs=null, and we might find clustered variants to put them in.

## Arising tasks (to be done in EVA-2033)

- make getAllByAccession
- remove uk.ac.ebi.eva.accession.core.service.nonhuman.eva.SubmittedVariantAccessioningDatabaseService.findByHashedMessageIn
- review if it makes sense to add getByIdentifying fields to the service
- last inactive in clustered service
- correct getVariantType
- others explained in the ticket.


eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteredVariantMergingPolicy.java
```diff
@@ -0,0 +1,48 @@
...
+            this.accessionToKeep = accessionToKeep;
+            this.accessionToBeMerged = accessionToBeMerged;
+        }
+    }
+
+    /**
+     * At the moment, the priority is just to keep the oldest accession, but other policies might apply, like choosing
+     * the accession with clinical relevance (which we don't store at the moment of writing).
+     *
+     * @see <a href="https://ncbijira.ncbi.nlm.nih.gov/secure/attachment/95534/95534_VAR-RSIDAssignment-100220-1039-34.pdf">
```
sundarvenkata-EBI, line 37: This is a minor thing but can we publish this document in Confluence (just in case NCBI takes it down in the future) and link to that document instead?

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
```diff
@@ -115,14 +148,176 @@ private VariantType getVariantType(String reference, String alternate) {
...
+                                  getClusteredVariantCollection(prioritised.accessionToBeMerged));
+
+        // Update submitted variants linked to the clustered variant we just merged.
+        // This has to happen for both EVA and dbsnp SS because previous cross merges might have happened.
+        updateSubmittedVariants(prioritised, SubmittedVariantEntity.class, SubmittedVariantOperationEntity.class);
+        updateSubmittedVariants(prioritised, DbsnpSubmittedVariantEntity.class,
+                                DbsnpSubmittedVariantOperationEntity.class);
+    }
+
+    private Class<? extends ClusteredVariantEntity> getClusteredVariantCollection(Long accession) {
```
sundarvenkata-EBI, line 186: May be this is already part of polishing but I feel that this and the following function should belong in a separate class since they represent a business rule

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
```diff
@@ -115,14 +148,176 @@ private VariantType getVariantType(String reference, String alternate) {
...
+
+        Query queryClustered = query(where(ACCESSION_KEY).is(prioritised.accessionToBeMerged));
+        List<? extends ClusteredVariantEntity> cvToMerge =
+                mongoTemplate.find(queryClustered, getClusteredVariantCollection(prioritised.accessionToBeMerged));
+
+        List<? extends ClusteredVariantEntity> cvToKeep =
+                mongoTemplate.find(query(where(ACCESSION_KEY).is(prioritised.accessionToKeep)),
+                                   getClusteredVariantCollection(prioritised.accessionToKeep));
+
+        if (isMultimap(cvToMerge) || isMultimap(cvToKeep)) {
```
sundarvenkata-EBI, line 162: Should we also ensure that this policy of not merging or "merging into" overweight SNP is conformed to in the **historic data** in dbsnpCVE/dbsnpCVOE collections? 

Consider the following scenario:
**Historic data (before the current policy)**
dbSNP_RS1 was merged into dbSNP_RS0 - dbSNP_RS0 is an overweight SNP - both RS are in ASM1
Therefore, dbSNP_RS1 is in dbSNPCVOE and dbSNP_RS0 is in dbSNPCVE

**New data**
SS3 submitted to ASM1 - when clustering SS3, dbsnp_RS0  appears to be a merge target based on SS3's locus but we don't pursue the RS assignment because it is overweight. So we proceed to issue a new RS ID for SS3. 
Here, we are paying a price for a historic mistake since the correct course of action would have been to assign dbSNP_RS1 as the RS for SS3.

eva-accession-clustering/src/test/java/uk/ac/ebi/eva/accession/clustering/batch/io/clustering_writer/MergeAccessionClusteringWriterTest.java
```diff
@@ -0,0 +1,505 @@
...
+    }
+
+    @After
+    public void tearDown() {
+        mongoTemplate.getDb().drop();
+    }
+
+    @Test
+    @DirtiesContext
+    public void merge_eva_clustered_accession() throws Exception {
```
sundarvenkata-EBI, line 147: Any reason we are eschewing Camel casing in this test class?

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
jmmut: I don't really understand why business rules should be treated differently than the rest of the code, but I see the value of extracting it to it's own class, maybe even in eva-accession-core. I'll add a note in the ticket for polishing.

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
jmmut: thinking a bit more about this example, that's not exactly what would happen. Bear in mind that there are 2 things that can happen in the clustering pipeline: clustering (add a rs to a ss if it is unclustered) and RS merging (when an ss is already clustered in rs1, but the locus after remapping makes collision with rs2).

In your scenario, it would just cluster SS3 under RS0, but no further merging will happen. Although technically this doesn't go against the policy of "not merging into overweight", arguably it is kind of wrong too. However, if we ever split those overweight RS, no unmerging is needed,  (other than the dbsnp one that already happened during the import), just change the RS when the locus is different. I agree on discussing this in the next meeting.

In a slightly different scenario, if SS3 had RS3 (instead of being unclustered), nothing would be changed. SS3 would keep RS3, even though RS3 is not in ASM1 (also, it can't be in ASM1 because it would make collision with RS0).

eva-accession-clustering/src/test/java/uk/ac/ebi/eva/accession/clustering/batch/io/clustering_writer/MergeAccessionClusteringWriterTest.java
jmmut: I tried to stick with camelCase but it was too hard to read. The alternative to underscores are more comments (that we won't read and won't update), so I think this is the least bad option.

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
```diff
@@ -115,14 +148,176 @@ private VariantType getVariantType(String reference, String alternate) {
...
+        operation.fill(EventType.MERGED, originalAccession, clusteredVariantMergedInto, reason,
+                       Collections.singletonList(inactiveEntity));
+        return operation;
+    }
+
+    /**
+     * This function updates the clustered variant accession (rs) of submitted variants when the rs makes a
+     * collision with another rs and they have to be merged.
+     *
+     * Note there is a similar scenario that is handled in another function: when a clustered variant accession (rs)
```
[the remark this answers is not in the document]
jmmut: renamed the second function to `clusterSubmittedVariants()` and removed extra comments.

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteredVariantMergingPolicy.java
jmmut: "RS ID merging/splitting policy", linked under "accessioning" in the main EVA page.

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
```diff
@@ -115,14 +148,176 @@ private VariantType getVariantType(String reference, String alternate) {
         return variantType;
     }
 
+    private void merge(Long providedAccession, String hash, Long accessionInDatabase) {
+        Priority prioritised = prioritise(providedAccession, accessionInDatabase);
+
+        Query queryClustered = query(where(ACCESSION_KEY).is(prioritised.accessionToBeMerged));
+        List<? extends ClusteredVariantEntity> cvToMerge =
```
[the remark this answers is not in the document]
jmmut: the abbreviation was more useful in previous versions of this code. renamed.

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
```diff
@@ -115,14 +148,176 @@ private VariantType getVariantType(String reference, String alternate) {
...
+
+    private ClusteredVariantOperationEntity buildClusteredOperation(ClusteredVariantEntity originalClusteredVariant,
+                                                                    Long clusteredVariantMergedInto) {
+        ClusteredVariantInactiveEntity inactiveEntity = new ClusteredVariantInactiveEntity(originalClusteredVariant);
+
+        Long originalAccession = originalClusteredVariant.getAccession();
+        String reason = "Original rs" + originalAccession + " was merged into rs" + clusteredVariantMergedInto + ".";
+
+        ClusteredVariantOperationEntity operation = new ClusteredVariantOperationEntity();
+        operation.fill(EventType.MERGED, originalAccession, clusteredVariantMergedInto, reason,
```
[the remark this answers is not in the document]
jmmut: I agree. I think we use consistently original/mergedInto across eva-accession, so I think that's good enough.

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
sundarvenkata-EBI: You are right, the terminology in my initial description does concern clustering. Nevertheless, IMO, RS0 should not be considered either as a merge target or as a RS assignment candidate.

eva-accession-clustering/src/test/java/uk/ac/ebi/eva/accession/clustering/batch/io/clustering_writer/ReuseAccessionClusteringWriterTest.java
```diff
@@ -0,0 +1,245 @@
...
+        ClusteredVariantEntity cve1 = createClusteredVariantEntity(asm1, existingRs);
+        mongoTemplate.insert(cve1);
+        assertEquals(1, mongoTemplate.count(new Query(), ClusteredVariantEntity.class));
+
+        // keep the RS that a submitted variant already has (RS=existingRs)
+        SubmittedVariantEntity sveClustered = createSubmittedVariantEntity(asm1, existingRs, 50L);
+        clusteringWriter.write(Collections.singletonList(sveClustered));
+        assertEquals(1, mongoTemplate.count(new Query(), ClusteredVariantEntity.class));
+
+        // assign an existing RS to a different submitted variant (that has rs=null)
```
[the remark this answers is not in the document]
jmmut: this seriously getting confusing. In this test we are giving the submitted variant directly to the clusteringWriter without storing it the DB, which means there are no submitted variants to update. I'm thinking that we should put serious effort in documenting what the pipeline does in different scenarios.

PR #225, opened 2020-07-22
EVA-2080 Add job to the release pipeline to separately output multi-map SNPs

jmmut:
Note that the VCF for multimap will contain in the header only the contigs that apply to that file.


eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/batch/processors/VariantToVariantContextProcessor.java
```diff
@@ -96,6 +97,7 @@ private String getSequenceName(String contig) {
                 case VARIANT_CLASS_KEY:
```
andresfsilva, line 97: No part of the new code but I think it worth correcting this. When the attribute key is VARIANT_CLASS_KEY it will add the MAPPING_WEIGHT_KEY as we don't have a break for the empty cases

jmmut: as discussed offline, the code is actually fine

PR #232, opened 2020-08-25
EVA-2127 Change multimap criteria to rely on the mapping weight

andresfsilva:


eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
```diff
@@ -224,12 +222,12 @@ private boolean isEvaClusteredAccession(Long accession) {
...
-                                                  .map(IClusteredVariant::getAssemblyAccession)
-                                                  .collect(Collectors.toSet())
-                                                  .size();
-        return assembliesCount < clusteredVariantLoci.size();
+    private boolean isMultimap(List<? extends IClusteredVariant> clusteredVariants) {
+        return clusteredVariants.stream().anyMatch(cv -> cv.getMapWeight() != null && cv.getMapWeight() > 1);
+    }
+
+    private boolean isMultimap(IClusteredVariant clusteredVariant) {
+        return isMultimap(Collections.singletonList(clusteredVariant));
```
tcezard, line 230: I'm not sure why there are two functions: one for a list and one for a single item. The only used one I can see is for the single item. I would suggest removing the one taking a list.  
```suggestion
    private boolean isMultimap(IClusteredVariant clusteredVariant) {
        return clusteredVariant.getMapWeight() != null && clusteredVariant.getMapWeight() > 1;
```

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
```diff
@@ -302,20 +300,11 @@ private SubmittedVariantOperationEntity buildSubmittedOperation(SubmittedVariant
...
-            if (isMultimap(allByAccession.stream().map(AccessionWrapper::getData).collect(Collectors.toList()))) {
-                multimapHashes.add(hashAndAccession.getKey());
-            }
-        }
-        multimapHashes.forEach(assignedAccessions::remove);
+    private void excludeMultimaps(List<GetOrCreateAccessionWrapper<IClusteredVariant, String, Long>> accessionWrappers) {
+        List<GetOrCreateAccessionWrapper<IClusteredVariant, String, Long>> accessionsNoMultimap =
+                accessionWrappers.stream().filter(x -> !isMultimap(x.getData())).collect(Collectors.toList());
+
+        accessionsNoMultimap.forEach(x -> assignedAccessions.put(x.getHash(), x.getAccession()));
```
tcezard, line 307: I'm not sure I expect this statement to be part of the `excludeMultimaps` function, It might just be me though.

andresfsilva: I was not sure either about that but I agree is better not to include it here. Changed the method, was thinking about just removing this method and doing the filtering directly in `getOrCreateClusteredVariantAccessions` but is probably clearer to keep the method also because of the comment in the header of the `excludeMultimaps` method

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
andresfsilva: The method that receives the list is used in the `merge` method. I could keep only one if you think this is creating confusion but it would be better to keep the one that takes the list, otherwise we would need to include loops when calling it. My preference is to keep both but let me know if you still think this makes the code less clear.

eva-accession-clustering/src/test/java/uk/ac/ebi/eva/accession/clustering/batch/io/clustering_writer/IssueAccessionClusteringWriterTest.java
```diff
@@ -243,22 +244,41 @@ private void assertSubmittedVariantsOperationInserted(){
 
     @Test
     @DirtiesContext
-//    @Ignore
-    public void do_not_cluster_eva_submitted_variant_into_a_dbsnp_multimap_clustered_variant() throws Exception {
+    public void cluster_eva_submitted_variant_into_a_dbsnp_not_multimap_clustered_variant() throws Exception {
```
sundarvenkata-EBI, line 247:
```suggestion
    public void cluster_eva_submitted_variant_into_a_dbsnp_non_multimap_clustered_variant() throws Exception {
```

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/model/eva/ClusteredVariantEntity.java
```diff
@@ -59,6 +61,16 @@ public ClusteredVariantEntity(Long accession, String hashedMessage, IClusteredVa
              model.getStart(), model.getType(), model.isValidated(), model.getCreatedDate(), version);
     }
 
+    /**
+     * This constructor should only be used when the mapping weight is required
+     */
+    public ClusteredVariantEntity(Long accession, String hashedMessage, IClusteredVariant model, int version,
+                                  Integer mapWeight) {
```
jmmut, line 68: agreed to put this parameter in the other constructor.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/model/eva/ClusteredVariantEntity.java
```diff
@@ -112,6 +125,11 @@ public Boolean isValidated() {
         return validated == null ? DEFAULT_VALIDATED : validated;
     }
 
+    @Override
+    public Integer getMapWeight() {
+        return mapWeight;
```
jmmut, line 130: agreed to put the same behaviour as in the validated flag. also agreed to add a test in eva-accession-core to check that the default value (mapWeight == 1) is not written in mongo, but it's ok when retrieving.

jmmut (changes requested):
Also, update the outdated comment at the top of the ClusteringWriter class, to explain that what takes priority is the mapWeight, not being multimap. Also, mention that the reason is that the check for multimap is too complicated because it should look into the operations collection and we lost some operations during the import (deprecation pipeline). Suggested wording:


 * Some edge cases take into account if a clustered variant is multimap. The definition of multimap variants that this
 * class uses is "clustered variants whose mapWeight is 2 or greater". Another definition is "clustered variants
 *  whose accession maps several times in the same assembly". Altough both definitions *should* yield the same 
 * set of variants, the check for the second definition is less efficient and less accurate: both the active and the
 * deprecated/merged collections should be queried, and we lost clusteredVariantOperations during the 
 * deprecation pipeline in the dbsnp import due to a bug.

sundarvenkata-EBI (approved):

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
tcezard: Yes you are right. I missed that. The code makes sense as it is.

tcezard (approved):

jmmut (approved):

PR #239, opened 2020-09-21
EVA-2122 Create a script that will handle release of species that did not change since last release

jmmut:
Note that these scripts are in eva-accession/eva-accession-release/src/main/resources. Maybe this should be in eva-accession/eva-accession-release-automation/, and if so, this PR is a good time to move the scripts there.


eva-accession-release/src/main/resources/create_hardlinks_for_unchanged_species.py
```diff
@@ -0,0 +1,74 @@
+
```
sundarvenkata-EBI, line 1: Would suggest moving to **eva-accession-release-automation** folder

eva-accession-release/src/main/resources/create_hardlinks_for_unchanged_species.py
```diff
@@ -0,0 +1,74 @@
...
+import glob
+import logging
+import subprocess
+import os
+from argparse import ArgumentParser
+from __init__ import *
+import create_assembly_name_symlinks
+
+
+species_name_mapping_file = 'species_name_mapping.tsv'
```
sundarvenkata-EBI, line 12: Would suggest using the [release_species_inventory table](https://www.ebi.ac.uk/panda/jira/browse/EVA-2121?focusedCommentId=337750&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-337750)

jmmut: agreed offline to not use the table for now.

sundarvenkata-EBI (approved):

PR #241, opened 2020-09-22
EVA-2158 Add dependency between bsub commands so they are run in series

andresfsilva:


eva-accession-clustering-automation/cluster_multiple_assemblies.py
```diff
@@ -44,6 +47,10 @@ def generate_bsub_command(assembly_accession, properties_path, clustering_artifa
     return command
 
 
+def get_job_name(assembly_accession):
+    return 'cluster_{assembly_accession}'.format(assembly_accession=assembly_accession)
```
jmmut, line 51: we have the option here of putting a timestamp in the job name. That way, if a run fails, the next run will be using a different set of dependencies.

In the past we have had problems when submitting different jobs with the same name, if they were submitted too fast (right after the previous one failed).

eva-accession-clustering-automation/cluster_multiple_assemblies.py
```diff
@@ -34,8 +34,11 @@ def generate_bsub_command(assembly_accession, properties_path, clustering_artifa
     if memory:
         memory_amount = memory
 
-    command = 'bsub -J {job_name} -o {log_file} -e {error_file} -M {memory_amount} -R "rusage[mem={memory_amount}]" ' \
-              'java -jar {clustering_artifact} --spring.config.location=file:{properties_path}'\
+    command = 'bsub '
+    if dependency:
+        command += '-w {dependency} '.format(dependency=dependency)
```
jmmut, line 39: you are mixing `+=` concatenation with newline concatenation (`' ' \\n ' '`) and  `.format`ting. I suggest a more consistent approach.

Also, not sure if you know you can do this:
```suggestion
        command += '-w {} '.format(dependency)
```
which may or may not be clearer, depending on the particular code.

jmmut (approved):
this looks like it works, but the changes I suggest are probably worth to do sooner rather than later.

eva-accession-clustering-automation/cluster_multiple_assemblies.py
andresfsilva: Good point, I just changed it to include the timestamp in the job name

eva-accession-clustering-automation/cluster_multiple_assemblies.py
andresfsilva: @jmmut Please let me know if you think the new approach looks better for you

tcezard (approved):

PR #255, opened 2021-04-13
Fix mongo error gitlab deployment and upgrade to variation commons 0.7.6

andresfsilva:
- The property `eva.mongo.host.test` is needed in all the tests that use mongo because gitlab needs the host to be `mongo` instead of `localhost` which is the default
- Included maven filtering for the properties files so the variable is translated into its value
- Bumper version of `variation-commons` to `0.7.6` so it includes the Chr fix
- Resumability tests from `EvaAccessionJobLauncherCommandLineRunnerTest.java` were ignored and will be revisit in a subsequent [ticket](https://www.ebi.ac.uk/panda/jira/browse/EVA-2392)


apriltuesday (approved):
Thanks Andres!

sundarvenkata-EBI (approved):

PR #256, opened 2021-04-16
Overridde BatchConfigurer to be backwards compatible with Spring Boot 1 job repository

sundarvenkata-EBI:
See https://stackoverflow.com/a/58707589


andresfsilva (approved):
I guess this will also need to be introduced for other pipelines like `eva-pipeline` and `eva-accession-clustering`

sundarvenkata-EBI:
> I guess this will also need to be introduced for other pipelines like `eva-pipeline` and `eva-accession-clustering`

Good point! It is already there in eva-pipeline. I will create a separate JIRA ticket to check and apply this change in other pipelines.

PR #258, opened 2021-04-26
EVA-2377 Ingest remapped variants

andresfsilva:


eva-ingest-remapped-variants/pom.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tcezard, line 1: Is it possible at all to have the tool for extraction `eva-accession-remapping` and ingestion `eva-ingest-remapped-variants` being co-located ? 
Or at least being named in a consistent way? Happy for `eva-accession-remapping` to be renamed if required.

eva-ingest-remapped-variants/src/main/java/uk/ac/ebi/eva/ingest/remapped/configuration/batch/io/VcfReaderConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+    @StepScope
+    public ItemStreamReader<Variant> unwindingReader(VcfReader vcfReader) {
+        return new UnwindingItemStreamReader<>(vcfReader);
+    }
+
+    @Bean
+    public VcfReader vcfReader(InputParameters inputParameters) throws IOException {
+        File vcfFile = new File(inputParameters.getVcf());
+        AggregatedVcfLineMapper lineMapper = new AggregatedVcfLineMapper("dummyFile", "dummyStudy", Aggregation.BASIC,
+                                                                         null);
```
tcezard, line 51: Are these dummy entries supposed to be there ?

eva-ingest-remapped-variants/src/test/resources/input-files/vcf/vcf_out_with_header.vcf
```diff
@@ -0,0 +1,5 @@
+##fileformat=VCFv4.3
+##contig=<ID=chr2>
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO
+chr2	98	ss5000000000	C	CG	50	PASS	PROJECT=projectId_1
+chr2	1078	ss5000000001	A	G	50	PASS	PROJECT=projectId_1
```
tcezard, line 5: Can you add an RSid in the info field I think it should be read and populated as well.

eva-ingest-remapped-variants/src/main/java/uk/ac/ebi/eva/ingest/remapped/batch/processors/VariantToSubmittedVariantEntityRemappedProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+        this.remappedFrom = remappedFrom;
+        hashingFunction = new SubmittedVariantSummaryFunction().andThen(new SHA1HashingFunction());
+    }
+
+    @Override
+    public SubmittedVariantEntity process(Variant variant) throws Exception {
+        long accession = Long.parseLong(variant.getMainId().substring(2));
+        VariantSourceEntry sourceEntry = variant.getSourceEntries().iterator().next();
+
+        String projectAccession = sourceEntry.getAttribute(PROJECT_KEY);
```
tcezard, line 55: Need to retrieve the RSid as well if it exists.

tcezard (commented):
I'm wondering what happen in two cases:
- If the same SubmittedVariant get remapped twice and placed both time in the same location in the new genome (same hash) would the ingestion overwrite the database document? 
- If the same SubmittedVariant get remapped twice and placed in two different location in the new genome (different hash) would the ingestion pick it up or both document will be kept?

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/model/eva/SubmittedVariantEntity.java
```diff
@@ -64,9 +66,18 @@
...
+    private String remappedFrom;
+
     protected SubmittedVariantEntity() {
     }
 
+    //Constructor to be used to store remapped submitted variants
+    public SubmittedVariantEntity(Long accession, String hashedMessage, ISubmittedVariant model, int version,
+                                  String remappedFrom) {
+        this(accession, hashedMessage, model, version);
+        this.remappedFrom = remappedFrom;
```
sundarvenkata-EBI, line 78: This is more of an enhancement suggestion (that can be handled in a separate ticket). Should we also have a re-mapping metadata (unique ID for a given remap run, from/to assemblies, timestamp etc.,)  in 
a separate collection somewhere and refer to that unique ID from within SVE? We can still retain remappedFrom in SVE for the sake of querying efficiency. Such cross-references could come in handy for troubleshooting.

sundarvenkata-EBI (commented):
Looks good! Will review again after @tcezard's suggestions have been incorporated.

eva-ingest-remapped-variants/src/test/resources/ingest-remapped-variants.properties
```diff
@@ -0,0 +1,26 @@
...
+accessioning.instanceId=instance-01
+accessioning.submitted.categoryId=ss
+accessioning.clustered.categoryId=rs
+
+accessioning.monotonic.ss.blockSize=100000
+accessioning.monotonic.ss.blockStartValue=5000000000
+accessioning.monotonic.ss.nextBlockInterval=1000000000
+accessioning.monotonic.rs.blockSize=100000
+accessioning.monotonic.rs.blockStartValue=3000000000
+accessioning.monotonic.rs.nextBlockInterval=1000000000
```
sundarvenkata-EBI, line 15: Although unncessary, the "accessioning.*" properties might be a necessary evil to satisy Spring autowiring gods but @andresfsilva will take another look to see if these can be dispensed with.

eva-ingest-remapped-variants/pom.xml
andresfsilva: I renamed them to `eva-remapping-ingest` and `eva-remapping-get-source`

eva-ingest-remapped-variants/src/test/resources/input-files/vcf/vcf_out_with_header.vcf
andresfsilva: Done

eva-ingest-remapped-variants/src/main/java/uk/ac/ebi/eva/ingest/remapped/batch/processors/VariantToSubmittedVariantEntityRemappedProcessor.java
andresfsilva: Done

eva-ingest-remapped-variants/src/main/java/uk/ac/ebi/eva/ingest/remapped/configuration/batch/io/VcfReaderConfiguration.java
andresfsilva: The file and study field are requested in the line mapper because this is also used when populating the `variant warehouse`. Those fields goes into the fields `files.fid` and `files.sid` in the variant warehouse but since that is not relevant for this pipeline we are just passing a dummy string as they are required.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/model/eva/SubmittedVariantEntity.java
andresfsilva: I think it makes sense to implement it but I agree it is better to do it in another ticket, this getting too big in scope. Added this in ticket [EVA-2420](https://www.ebi.ac.uk/panda/jira/browse/EVA-2420)

eva-ingest-remapped-variants/src/test/resources/ingest-remapped-variants.properties
andresfsilva: I think this should be done but will require a new configuration. Added in ticket [EVA-420](https://www.ebi.ac.uk/panda/jira/browse/EVA-2420)

andresfsilva:
@tcezard I added tests for these two scenarios so we have clear the current behaviour and decide if we want to change it.

> I'm wondering what happen in two cases:
> 
> * If the same SubmittedVariant get remapped twice and placed both time in the same location in the new genome (same hash) would the ingestion overwrite the database document?

In this case mongo will throw a duplicate key exception and we will skip the variant and log the message as a warning. Once the listener of the pipeline is implemented it can be recorded in the counts

> * If the same SubmittedVariant get remapped twice and placed in two different location in the new genome (different hash) would the ingestion pick it up or both document will be kept?

Yes, both documents would be kept as they have two different hashes. We could introduce some validation if we want to make sure there can only be one variant per `ss id` and `assembly`? But I am not sure if that is really what we want. We can discuss that

eva-remapping-ingest/src/main/java/uk/ac/ebi/eva/remapping/ingest/batch/io/RemappedSubmittedVariantsWriter.java
```diff
@@ -0,0 +1,52 @@
...
+    public void write(List<? extends SubmittedVariantEntity> submittedVariantsRemapped) throws Exception {
+        try {
+            BulkOperations bulkOperations = mongoTemplate.bulkOps(BulkOperations.BulkMode.UNORDERED,
+                                                                  SubmittedVariantEntity.class,
+                                                                  BeanNames.SUBMITTED_VARIANT_ENTITY);
+            bulkOperations.insert(submittedVariantsRemapped);
+            bulkOperations.execute();
+        } catch (DuplicateKeyException exception) {
+            logger.warn(exception.toString());
+        }
```
tcezard, line 50: Will this exception prevent the loading of all the variants in the `List` ? Or will only the duplicated one?

tcezard (approved):
I think this looks good. Lets discuss the duplicates issue and potentially add it to [EVA-2420](https://www.ebi.ac.uk/panda/jira/browse/EVA-2420)

sundarvenkata-EBI (approved):

eva-remapping-ingest/src/main/java/uk/ac/ebi/eva/remapping/ingest/batch/io/RemappedSubmittedVariantsWriter.java
andresfsilva: No, it will write all the variants except for the duplicates that will be present in the exception

PR #261, opened 2021-05-07
EVA-1933 - Rectify pipelines with Spring Boot 1 job repo. compat.

sundarvenkata-EBI:


andresfsilva (commented):
We are not updating the configuration for 3 modules that use Spring Batch
- `eva-accession-import` (No used anymore)
- `eva-accession-deprecate` (Not used anymore)
- `eva-accession-import-dbsnp2` (Import human from dbsnp, not used at the moment be will probably be)

I think we should either update them so they are not inconsistent as we upgraded the repo to Spring 2 or we should consider archiving them if we are not going to use them anymore

sundarvenkata-EBI:
> We are not updating the configuration for 3 modules that use Spring Batch
> 
> * `eva-accession-import` (No used anymore)
> * `eva-accession-deprecate` (Not used anymore)
> * `eva-accession-import-dbsnp2` (Import human from dbsnp, not used at the moment be will probably be)
> 
> I think we should either update them so they are not inconsistent as we upgraded the repo to Spring 2 or we should consider archiving them if we are not going to use them anymore

Archiving these is a longer-term discussion. Agree with you on consistency so I have updated the modules.

andresfsilva (approved):

PR #262, opened 2021-05-10
EVA-2424 Remove filter that excludes clustered variants from reader

andresfsilva:
- When clustering we should take into account all variants, even if they are already clustered
- When a submitted variant is already clustered and is remapped (`remappedFrom` fields is not null) the pipeline checks if a merge is required


tcezard (approved):
That's probably for another ticket but I think we should remove the path that allow the variants to be read from VCF.

PR #263, opened 2021-05-11
EVA-2420 Improve remapped variants ingestion pipeline

andresfsilva:
- [x] Add listener to have counts
- [x] Remove accession properties
- [x] Add application properties file
- [x] Add test for job

### DIFF
--- a/eva-remapping-ingest/src/main/resources/application.properties
+++ b/eva-remapping-ingest/src/main/resources/application.properties
@@ -1,0 +1,20 @@
+spring.batch.job.names=INGEST_REMAPPED_VARIANTS_FROM_VCF_JOB
+
+parameters.vcf=
+parameters.assemblyAccession=
+parameters.remappedFrom=
+parameters.taxonomyAccession=
+parameters.chunkSize=100
+
+spring.data.mongodb.database=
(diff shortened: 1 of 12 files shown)


eva-remapping-ingest/src/main/resources/application.properties
```diff
@@ -0,0 +1,20 @@
+spring.batch.job.names=INGEST_REMAPPED_VARIANTS_FROM_VCF_JOB
```
tcezard, line 1: Does it need to be called `application.properties` ?

tcezard (approved):

eva-remapping-ingest/src/main/resources/application.properties
andresfsilva: It does not need to be called `application.properties` but that is the default file Spring looks for. If we want to use a different name it is possible but we will need to specify it via environment properties `spring.config.name` or `spring.config.location`. See [here](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-files) for reference.

sundarvenkata-EBI (approved):

PR #271, opened 2021-06-14
Change taxonomy key to be the same in the remapping extract and ingest

andresfsilva:


apriltuesday:
@andresfsilva I completely forgot about this one, we should try to merge this as well!  If it needs a JIRA ticket you can use EVA-2463.

andresfsilva:
> @andresfsilva I completely forgot about this one, we should try to merge this as well! If it needs a JIRA ticket you can use EVA-2463.

@apriltuesday I included these changes as part of EVA-2470. The commits are in PR #272. I will close this PR so it does not cause confusion. Thanks!

PR #276, opened 2021-07-12
EVA-2491: remove fasta index if outdated

apriltuesday:


tcezard (approved):
This assumes that the subsequent step will recreate the `fai` file which I believe is correct.

sundarvenkata-EBI (approved):

PR #277, opened 2021-07-19
EVA-2486 / EVA-2190 - Drive clustering from unified clustering/release tracking table

apriltuesday:
* Drive clustering from unified clustering/release tracking table
* Split clustering from mongo and clustering from vcf functionality
* Use nextflow for clustering from mongo
* Include instance parameter in properties file (both mongo and vcf) and restructure output directory (mongo only)


eva-accession-clustering-automation/clustering_automation/cluster_from_mongo.py
```diff
@@ -0,0 +1,126 @@
...
+        pipeline._add_new_process(NextFlowProcess(
+            process_name=f'cluster_{suffix}',
+            command_to_run=f'java -jar {clustering_artifact} --spring.config.location=file:{properties_path}',
+            process_directives={'memory': f'{memory} MB',
+                                'clusterOptions': (f'-o {output_directory}/cluster_{timestamp}.log '
+                                                   f'-e {output_directory}/cluster_{timestamp}.err')}
+        ))
+        pipeline.add_process(
+            process_name=f'end_{suffix}',
+            command_to_run=status_update_template.format(status='Completed')  # TODO: how to choose completed/failed?
```
tcezard, line 86: That would be one of the reason why it might not be such a good idea to have the status update in different nextflow steps

eva-accession-clustering-automation/clustering_automation/create_clustering_properties.py
```diff
@@ -68,11 +68,11 @@ def get_job_name(source):
...
 
 
-def add_accessioning_properties(properties_file):
-    properties_file.write("""
+def add_accessioning_properties(properties_file, instance):
+    properties_file.write(f"""
 parameters.chunkSize=100
 
-accessioning.instanceId=instance-01
+accessioning.instanceId=instance-{instance}
```
tcezard, line 75:
```suggestion
accessioning.instanceId=instance-{instance:02d}
```

tcezard (approved):

eva-accession-clustering-automation/clustering_automation/cluster_from_mongo.py
apriltuesday: I wasn't quite sure which way to go... the [release automation](https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-release-automation/run_release_in_embassy/run_release_for_species.py#L56) right now does something similar to this (status update in nextflow, doesn't mark things as failed), while [remapping automation](https://github.com/apriltuesday/eva-tools/blob/master/variant-remapping-automation/remapping_automation.py#L201) does mark failures but uses a separate nextflow pipeline for each assembly.

tcezard: Yes it is something we will have to tackle when we harmonise the three automation in the same pipeline.

apriltuesday: I could refactor and do the status updates for each assembly outside nextflow, it will be a little silly to have a pipeline with just one step but if we do in fact add QA steps then it could make sense.

andresfsilva (approved):

eva-accession-clustering-automation/clustering_automation/cluster_from_mongo.py
apriltuesday: We discussed this a bit this morning, but agreed it would make more sense to make everything consistent once we bring the automation scripts together.

PR #279, opened 2021-07-25
EVA-2515 Update release automation to use new tracking table

andresfsilva:


eva-accession-release-automation/README.md
```diff
@@ -23,7 +23,6 @@ release-version:
 release-folder:
 public-ftp-release-base-folder:
 release-species-inventory-table:
-release-progress-table:
```
apriltuesday, line 26: I actually think `release-progress-table` is a better name, but I don't think it matters because we'll probably unify this with the [clustering automation's config](https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-clustering-automation/clustering_automation/cluster_from_mongo.py#L54) soon anyway.

apriltuesday (approved):

tcezard (approved):

PR #282, opened 2021-08-03
EVA-2529: Process variant with RS id first

nitin-ebi:


eva-accession-clustering/src/test/java/uk/ac/ebi/eva/accession/clustering/runner/ClusteringCommandLineRunnerTest.java
```diff
@@ -86,6 +87,9 @@
     @Autowired
     private ClusteringCommandLineRunner runner;
 
+    @Autowired
+    private ClusteringCommandLineRunner mongoRunner;
```
andresfsilva, line 91: Can you reuse the existing runner?

nitin-ebi: updated

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
```diff
@@ -139,6 +145,60 @@ private void getOrCreateClusteredVariantAccessions(List<? extends SubmittedVaria
...
+     */
+    private List<IClusteredVariant> processClusteredVariantsWhereNoRSExists(List<? extends SubmittedVariantEntity> submittedVariantEntities) {
+        List<SubmittedVariantEntity> submittedVariantWithRS = submittedVariantEntities.stream()
+                .filter(v -> v.getClusteredVariantAccession() != null)
+                .collect(Collectors.toList());
+        List<? extends SubmittedVariantEntity> submittedVariantsNoRSExists = getSubmittedVariantsNoRSExists(submittedVariantWithRS);
+        List<ClusteredVariantEntity> clusteredVariantEntityNoRSExists = submittedVariantsNoRSExists.stream()
+                .map(this::toClusteredVariantEntity)
+                .collect(Collectors.toList());
+        mongoTemplate.insert(clusteredVariantEntityNoRSExists, ClusteredVariantEntity.class);
```
andresfsilva, line 174: This should only be done for variants that have been remapped so it would be better to validate that. 
Variants that are clustered and not remapped should always have a corresponding/existing RS id but still, to be safe I suggest adding the check.

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
```diff
@@ -139,6 +145,60 @@ private void getOrCreateClusteredVariantAccessions(List<? extends SubmittedVaria
...
+     * No RS id exist for the HASH ASM2_Chr1_2_A_T,	but it has an existing rs id of 300, which it should continue to use.
+     * It should not generate a new rs id for this situation as they will get merged right after generation and will be wasted
+     *
+     *         ClusteredVariantEntity (after operation)
+     *          RS	HASH	        ASM	    POS	CONTIG	TYPE
+     *          300	ASM1_Chr1_1_SNV	ASM1	1	Chr1	SNV
+     *          300	ASM2_Chr1_2_SNV	ASM1	1	Chr1	SNV
+     */
+    private List<IClusteredVariant> processClusteredVariantsWhereNoRSExists(List<? extends SubmittedVariantEntity> submittedVariantEntities) {
+        List<SubmittedVariantEntity> submittedVariantWithRS = submittedVariantEntities.stream()
```
andresfsilva, line 167: I am not sure about this part. We separated the mongo readers to process first clustered and then unclustered variants so the filter is already happening at the database level but we are checking it again here.

andresfsilva (changes requested):

eva-accession-clustering/src/test/java/uk/ac/ebi/eva/accession/clustering/configuration/batch/steps/ClusteringVariantStepConfigurationTest.java
```diff
@@ -122,7 +122,7 @@ public void stepFromVcf() {
         assertEquals(BatchStatus.COMPLETED, jobExecution.getStatus());
 
         assertEquals(5, mongoTemplate.getCollection(SUBMITTED_VARIANT_COLLECTION).countDocuments());
-        assertClusteredVariantsCreated(Arrays.asList(3000000000L, 3000000001L, 3000000002L));
+        assertClusteredVariantsCreated(Arrays.asList(30L, 3000000000L, 3000000001L, 3000000002L));
```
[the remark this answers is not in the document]
nitin-ebi: reverted to original

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
```diff
@@ -139,6 +145,60 @@ private void getOrCreateClusteredVariantAccessions(List<? extends SubmittedVaria
...
+                .map(d -> new GetOrCreateAccessionWrapper<>(d.getAccession(), d.getHash(), d.getData(), false))
+                .collect(Collectors.toList());
+        List<IClusteredVariant> clusteredVariantsRSExists = accessionWrappers.stream()
+                .map(v -> v.getData())
+                .collect(Collectors.toList());
+
+        return submittedVariantWithRS.stream()
+                .filter(sub -> !clusteredVariantsRSExists.contains(toClusteredVariant(sub)))
+                .collect(Collectors.toList());
+    }
```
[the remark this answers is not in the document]
nitin-ebi: updated.
for last filtering I am using 
  return submittedVariantWithRS.stream()
                .filter(sve->!existingClusteredVariantsHashes.contains(**toClusteredVariantEntity(sve)**.getHashedMessage()))

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
nitin-ebi: added a filter to take only remapped submittedVariants

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
nitin-ebi: Again the writer is same, so the same method will get called in case of both the steps. This method does not make sense in case of step 2 as all variants will be non clustered so should get filtered out. 
we can take the same approach as reader and create 2 different beans for writers with one Boolean field, that will determine whether we need to call this method or not.

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringMongoReader.java
```diff
@@ -58,13 +59,16 @@
...
+
     public ClusteringMongoReader(MongoClient mongoClient, String database, MongoTemplate mongoTemplate,
-                                 String assembly, int chunkSize) {
+                                 String assembly, int chunkSize, boolean isClusteredVariantsMongoReader) {
         this.mongoClient = mongoClient;
         this.database = database;
         this.mongoTemplate = mongoTemplate;
         this.assembly = assembly;
         this.chunkSize = chunkSize;
+        this.isClusteredVariantsMongoReader = isClusteredVariantsMongoReader;
```
sundarvenkata-EBI, line 71: Since this particular variable name isn't self-explanatory, suggest adding a comment as to why we need this.

nitin-ebi: changed the name to **readOnlyClusteredVariants** to make it more clear and also added comment

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
andresfsilva: I would rather that instead and not filtering the variants again. That way is clear in which cases it should not be called

sundarvenkata-EBI (approved):

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
nitin-ebi: updated

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringMongoReader.java
```diff
@@ -80,8 +85,8 @@ private SubmittedVariantEntity getSubmittedVariantEntity(Document notClusteredSu
     public void open(ExecutionContext executionContext) throws ItemStreamException {
         MongoDatabase db = mongoClient.getDatabase(database);
         MongoCollection<Document> collection = db.getCollection(SUBMITTED_VARIANT_ENTITY);
-
-        Bson query = Filters.and(Filters.in(ASSEMBLY_FIELD, assembly));
+        Bson query = Filters.and(Filters.in(ASSEMBLY_FIELD, assembly),
+                Filters.exists(CLUSTERED_VARIANT_ACCESSION_FIELD, readOnlyClusteredVariants));
```
tcezard, line 89: I know this was not specified in the scope but I think we should also filter on the existence of the  `remappedFrom` field because we only want to cluster the remapped variant that have an RSid not the one that were created in this assembly.

nitin-ebi: we have the filtering on **"remappedFrom"** field in java. 
If we want to put the filter at the DB level, we will have to separate out the query because this query is same for both mongo readers (clustered and non clustered reader). If we put it here, even in case of non-clustered it will try to search for that field.

tcezard: I see.

tcezard (approved):

PR #296, opened 2021-09-15
EVA-2237 Add remapping fields to ws response

andresfsilva:


tcezard (approved):
This looks good although I'm nopt sure we want to expose the `remappingId` to the public. It will not be useful and might be confusing to external users

sundarvenkata-EBI (approved):

PR #310, opened 2021-11-15
EVA-2634 - Run the nextflow inside the assembly directory 

tcezard:
so multiple process don't share the same nextflow work directory and logs

### DIFF
--- a/eva-accession-clustering-automation/clustering_automation/cluster_from_mongo.py
+++ b/eva-accession-clustering-automation/clustering_automation/cluster_from_mongo.py
@@ -56,8 +56,9 @@ def generate_linear_pipeline(taxonomy_id, scientific_name, assembly_list, common
     clustering_tracking_table = common_properties['clustering-release-tracker']
 
     pipeline = LinearNextFlowPipeline()
+    species_directory = os.path.join(clustering_folder, f"{scientific_name.lower().replace(' ', '_')}_{taxonomy_id}")
     for assembly in assembly_list:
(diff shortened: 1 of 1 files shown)


apriltuesday (approved):

eva-accession-clustering-automation/clustering_automation/cluster_from_mongo.py
```diff
@@ -92,15 +92,15 @@ def generate_linear_pipeline(taxonomy_id, scientific_name, assembly_list, common
...
+                                'clusterOptions': f"{process_directives_for_java_pipelines['clusterOptions']}"
+                                                  f" -g /accession/{instance} "}  # needed to serialize accessioning
         ))
         pipeline.add_process(
             process_name=f'end_{suffix}',
             command_to_run=status_update_template.format(status='Completed')  # TODO: how to choose completed/failed?
         )
         # TODO add QA process
-    return pipeline
+    return pipeline, output_directory
```
andresfsilva, line 103: The `output_directory` is different for every [assembly](https://github.com/EBIvariation/eva-accession/pull/310/files#diff-f5d03ec235616a63adf93f3730b453bd9e57bba3743c657ec38025e1a3e9302fL60) so we would be returning the output directory for the last assembly only but the linear pipeline might include multiple assemblies. is that ok?

tcezard: Good point. I changed this to the parent directory which is the species

andresfsilva (approved):

PR #316, opened 2021-11-30
EVA-2671: Retryable reads for clustering

apriltuesday:
* Sorts clustering reader queries by id, tracks current id and collection (i.e. dbSNP or EVA) in execution context
* Retries reads that fail with `MongoCursorNotFoundException` by re-querying Mongo with filter and retrying the read


eva-accession-clustering/src/test/java/uk/ac/ebi/eva/accession/clustering/configuration/batch/steps/ClusteringVariantStepConfigurationTest.java
```diff
@@ -246,7 +246,7 @@ private void assertGeneratedAccessions(String collectionName, String accessionFi
 
     private void assertSubmittedVariantsUpdated() {
         assertTrue(allSubmittedVariantsClustered());
-        List<Long> expectedAccessions = Arrays.asList(3000000000L, 3000000000L, 3000000001L, 3000000002L, 3000000003L);
+        List<Long> expectedAccessions = Arrays.asList(3000000000L, 3000000001L, 3000000002L, 3000000002L, 3000000003L);
```
apriltuesday, line 249: rsids assigned change slightly because of the change in the order that variants are read in

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringMongoReader.java
```diff
@@ -79,24 +116,42 @@ private SubmittedVariantEntity getSubmittedVariantEntity(Document notClusteredSu
 
     @Override
     public void open(ExecutionContext executionContext) throws ItemStreamException {
+        if (executionContext.containsKey(CURRENT_ID_KEY)) {
```
sundarvenkata-EBI, line 119: An idea for the future might be to check if this is persisted across multiple runs of a job. If so, we can leverage it to resume jobs from where we left off.

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringMongoReader.java
```diff
@@ -106,13 +161,15 @@ public void initializeReader() {
     private FindIterable<Document> getSubmittedVariants(Bson query, Class<?> entityClass) {
         return mongoTemplate.getCollection(mongoTemplate.getCollectionName(entityClass))
                             .find(query)
+                            .sort(Sorts.ascending(ID_FIELD))
```
sundarvenkata-EBI, line 164: Does adding a limit function with the chunkSize cause any issues here? If not, I will recommend adding it because I am not sure if Spring does that internally.

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringMongoReader.java
```diff
@@ -122,4 +179,24 @@ public void close() throws ItemStreamException {
...
+     */
+    class MongoReaderRetryPolicy extends SimpleRetryPolicy {
+
+        public MongoReaderRetryPolicy() {
+            super(MAX_RETRIES, Collections.singletonMap(MongoCursorNotFoundException.class, true));
+        }
+
+        @Override
+        public void registerThrowable(RetryContext context, Throwable throwable) {
+            if (canRetry(context) && context.getRetryCount() < getMaxAttempts()-1
```
sundarvenkata-EBI, line 195: Since we are not comparing against an array index, I am not sure why we have a n-1 here...

eva-accession-clustering/src/test/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringMongoReaderRetryTest.java
```diff
@@ -0,0 +1,158 @@
...
+        verify(mockCursor, times(4)).next();
+        // initializeReader() called once at the start and once during the retry
+        verify(nonClusteredVariantReader, times(2)).initializeReader();
+    }
+
+    @Test
+    public void readFailsWhenOtherExceptionThrownOnce() {
+        MongoCursor mockCursor = mock(MongoCursor.class);
+        when(mockCursor.hasNext()).thenReturn(true, true, true, true, false);
+        Document doc1 = new Document("id_", "1");
```
sundarvenkata-EBI, line 100: Are these "id_" as opposed to "_id" for a particular reason? The reader sort does seem to work with the correct attribute "_id"..

sundarvenkata-EBI (approved):
Well thought out! I am palpably excited to have this functionality in clustering!!

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringMongoReader.java
```diff
@@ -67,10 +89,25 @@ public ClusteringMongoReader(MongoTemplate mongoTemplate, String assembly, int c
...
+    public SubmittedVariantEntity doRead() {
         // Read from dbsnp collection first and subsequently EVA collection
-        Document nextElement = (dbsnpCursor != null && dbsnpCursor.hasNext()) ? dbsnpCursor.next() :
-                (evaCursor.hasNext() ? evaCursor.next() : null);
-        return (nextElement != null) ? getSubmittedVariantEntity(nextElement) : null;
+        Document nextElement = null;
+        if (dbsnpCursor != null && dbsnpCursor.hasNext()) {
+            currentCollection = DBSNP_COLLECTION;
+            nextElement = dbsnpCursor.next();
+        } else if (evaCursor.hasNext()) {
```
sundarvenkata-EBI, line 101: evaCursor can be null too. So suggest adding "evaCursor != null" here..

tcezard: `dbsnpCursor` can be `null` because we set it to null. I don't think `evaCursor` can be null

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringMongoReader.java
```diff
@@ -79,24 +116,42 @@ private SubmittedVariantEntity getSubmittedVariantEntity(Document notClusteredSu
...
+
     public void initializeReader() {
         Bson query = Filters.and(Filters.in(ASSEMBLY_FIELD, assembly),
                                  Filters.exists(CLUSTERED_VARIANT_ACCESSION_FIELD, readOnlyClusteredVariants));
+        Bson queryWithCurrentId = null;
+        if (currentId != null) {
+            queryWithCurrentId = Filters.and(query, Filters.gt(ID_FIELD, currentId));
+            logger.info("Issuing find in {} collection: {}", currentCollection, queryWithCurrentId);
+        }
         logger.info("Issuing find: {}", query);
```
tcezard, line 141: When you retry you would log twice. Maybe doing it only once with a generic log is better?

tcezard (approved):
Do you think we'll a see the CursorNotFound Error in the logs? 
I think we will want to continue to count how many time it occurs

eva-accession-clustering/src/test/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringMongoReaderRetryTest.java
apriltuesday: no reason except that I screwed it up... thanks!

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringMongoReader.java
apriltuesday: I think setting the cursor's batchSize should be enough, right?  Otherwise we'd have to call `initializeReader` for every chunk...

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringMongoReader.java
apriltuesday: It's a bit confusing actually, when we retry we might actually be making two different queries - for example, if `readOnlyClusteredVariants` is true and the failure occurred while reading the dbSNP collection, we'll query the dbSNP collection with the currentId filter but the EVA collection with the standard query.

I was being lazy about the logging but I'll change it to be more specific as to what we're querying in which collection, because I can definitely see us getting confused by it in the (near) future...

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringMongoReader.java
apriltuesday: The n-1 is so that we don't initialize unnecessarily after the last attempt fails, see e.g. [this test](https://github.com/EBIvariation/eva-accession/pull/316/files/825ca3c8c9843b16d9e272b7bb5ffa7d11209103#diff-658a169121fd1f14103fbfd338d50b93f547b93c82d7ec24c2a81fa06fd044d4R92).  I think it's because the `retryCount` is getting incremented in the superclass's `registerThrowable`.  I'll add a comment explaining this...

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringMongoReader.java
apriltuesday: I think Tim is right that `evaCursor` should never be null, but I guess adding the check doesn't hurt...

sundarvenkata-EBI (approved):

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringMongoReader.java
apriltuesday: I was wondering this as well, for jobs I ran in dev it's persisted as part of the step's execution context but not the job's execution context... I'm not sure if that's enough for Spring to use it to restart or whether we need to propagate it to the job level, but it would be cool to get this working!

I don't think it would add any overhead since Spring is already doing some context persistence, we're just adding a couple fields.  Others might know more about this though...

apriltuesday:
@tcezard Also added a [warning](https://github.com/EBIvariation/eva-accession/pull/316/files#diff-0dddd470681fc67682571c476a8060e7892fb795ef5ffde6080f4dc1ab54514bR207) for CursorNotFoundExceptions so we can continue to see them in logs, together with the new query being issued.

sundarvenkata-EBI (approved):

tcezard (approved):

PR #319, opened 2021-12-08
EVA-2684 - Fix/merge split handling idempotency

sundarvenkata-EBI:


eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/RSSplitWriter.java
```diff
@@ -140,6 +147,9 @@ private void issueNewRSForHashes(List<String> hashesThatShouldGetNewRS,
                 submittedVariantInactiveEntities
                 .stream()
                 .map(SubmittedVariantInactiveEntity::toSubmittedVariantEntity)
+                // If the split was processed multiple times due to multiple runs of the clustering pipeline
+                // ensure that the SS was not already assigned a new RS
+                .filter(ss -> !this.doesSSAlreadyHaveANewRS(ss))
```
tcezard, line 170: I'm still a bit confuse as to how this scenario would occur. 
If I run the Split detection multiple times we would only record this event once since it is now idempotent.
If I run the Split resolution multiple times the candidate would have been deleted from the operations so will not be processed multiple times.

I guess there might be a possibility that the candidate be processed and not deleted if there is a crash between the two operation but that seems unlikely.

apriltuesday: If I understand correctly, the deletion of candidates happens in a [separate step](https://github.com/EBIvariation/eva-accession/blob/016120895efe60fb0a8c25e50a6c4c1ed8a74091/eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/configuration/batch/jobs/ClusterUnclusteredVariantsJobConfiguration.java#L43), so a crash anytime during the split processing step could cause split candidates to be re-processed when the job is rerun.  So maybe your latter scenario isn't so unlikely.

apriltuesday (commented):
Mostly questions from me, otherwise this looks good!

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/RSSplitWriter.java
tcezard: I see.
Maybe we should put the deletion of candidate in the same step to make it more robust.

andresfsilva: I am not sure what the reason was to not delete the candidates immediately after processing/resolving them. Maybe something regarding multi-merges/splits or some other complex scenarios. I think we can move with the current implementation (having the deletion in a subsequent step)

andresfsilva (approved):
PR looks good to me. There are maybe improvements that could be made regarding how we are dealing with split/merge candidates in subsequent tickets

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/RSSplitWriter.java
sundarvenkata-EBI: If candidates are deleted while there is an active cursor iterating through those candidates in the candidate processing step (sort of "pulling the rug from underneath" the cursor), it can (and did) lead to undefined results. Hence the separation was done.

tcezard: This makes sense but we could still imagine marking these candidate for deletion instead of deleting them. This would ensure that they are treated in a more atomic fashion.
We can do that in a subsequent improvement.

PR #320, opened 2021-12-15
EVA-2704 Keep submittedVariantOperationEntity in memory instead of querying it every batch

andresfsilva:


tcezard (approved):
This looks good.

nitin-ebi (approved):

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
```diff
@@ -194,17 +213,8 @@ private void processClusteredRemappedVariants(List<? extends SubmittedVariantEnt
 
         String assembly = clusteredRemappedSubmittedVariants.get(0).getReferenceSequenceAccession();
```
tcezard, line 214: Do you need to get the assembly from the clustered variant now that you have it populated in the constructor ?

tcezard: Actually now that the assembly is a property you can avoid passing it around a lot of methods

PR #325, opened 2022-01-13
EVA-2655 - Fix issue in counts for submitted variant accessioning

apriltuesday:


tcezard (approved):

sundarvenkata-EBI:
I am ok with the reasoning behind this PR and the code. But I am wondering why this issue surfaced during this particular batch (batch 18) when we have had other batches that were processed in the past which could have posed a similar situation for this issue to arise.

apriltuesday:
@sundarvenkata-EBI I searched the accessioning logs for previous batches and didn't see any error...  But also, one reason it wouldn't show up very frequently is that the more thorough count check only happens when the number of variants and the number of accessions doesn't match, see [here](https://github.com/EBIvariation/eva-accession/pull/325/files#diff-b505f15a2709399f6ebbd4535c5de5816b0e2f10106aff4df210e0e11e29a105L71).  In this case this was triggered because there was _actually_ a duplicate variant, at which point the rest of the method ran into this bug and foundered, causing the pipeline to crash.

Thank you for asking about this, I hadn't even thought about this reasoning until just now!

sundarvenkata-EBI (approved):

PR #337, opened 2022-03-31
EVA-2805 Identifier web services contig alias integration

andresfsilva:


eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/contigalias/ContigAliasService.java
```diff
@@ -0,0 +1,209 @@
...
+                contig = contigAliasChromosome.getEnaSequenceName();
+                break;
+            case MD5_CHECKSUM:
+                contig = contigAliasChromosome.getMd5checksum();
+                break;
+            case TRUNC512_CHECKSUM:
+                contig = contigAliasChromosome.getTrunc512checksum();
+                break;
+            default:
+                contig = contigAliasChromosome.getGenbank();
```
tcezard, line 130: Just to be consistent, I would prefer if we switched all the usage of `Genbank` to `Insdc` and the `Name` to `GenbankSequenceName`

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/contigalias/ContigAliasNaming.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+public enum ContigAliasNaming {
+    GENBANK_SEQUENCE_NAME,
+    ENA_SEQUENCE_NAME,
+    INSDC,  // this is the same as GenBank,
+    REFSEQ,
+    UCSC,
+    MD5_CHECKSUM,
+    TRUNC512_CHECKSUM,
+    NO_REPLACEMENT  // do not use any particular naming, just keep whatever contig is provided
```
tcezard, line 31: Is there a way to have these term being lower cased and dot (.) separated? 
I'm trying to follow the standard from [identifiers.org|https://registry.identifiers.org/registry]

sundarvenkata-EBI: @tcezard  While it is not possible to have the Java identifiers conform to the convention, we can follow [this method](https://stackoverflow.com/a/3978690) to attach string labels to the Java identifiers. These string labels can be made to conform to the convention that you suggest.

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/contigalias/ContigAliasService.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    /**
+     * Query contig alias service to translate the contig to the desired naming convention
+     */
+    private String translateContig(String genbankContig, ContigAliasNaming contigAliasNaming) {
+        String url = contigAliasUrl + CONTIG_ALIAS_CHROMOSOMES_GENBANK_ENDPOINT + genbankContig;
+        ContigAliasResponse contigAliasResponse = restTemplate.getForObject(url, ContigAliasResponse.class);
+        if (contigAliasResponse == null || contigAliasResponse.getEmbedded() == null) {
+            throw new NoSuchElementException("Not data returned for " + url + " from the contig alias service");
```
sundarvenkata-EBI, line 83:
```suggestion
            throw new NoSuchElementException("No data returned for " + url + " from the contig alias service");
```

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/contigalias/ContigAliasNaming.java
apriltuesday: While we're talking about names, I'm not totally sold on the API parameter name `contigAliasNaming` unless it's standard terminology.  Something like `contigNamingConvention` seems more natural to me.

(This is purely from the user's perspective, internally we can call it whatever we like.)

apriltuesday: I did what @sundarvenkata-EBI suggested for the string labels, but I don't think it actually affects what the parameter value is in the API as Spring seems to use the name rather than the value of enum... :roll_eyes:  I tried a few other things including using a custom converter and didn't get that to work either, so I've left it for now.

Worth pointing out we don't do this for other enums in our API, e.g. choice of variant type, so at least we're consistent!

eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/contigalias/ContigAliasService.java
apriltuesday: This should be consistent across this PR now but not necessarily in contig-alias itself, the latter will probably be addressed in [EVA-2838](https://www.ebi.ac.uk/panda/jira/browse/EVA-2838).

sundarvenkata-EBI (approved):

nitin-ebi (approved):

PR #344, opened 2022-05-20
Ignore clustered variant from both sides

tcezard:


sundarvenkata-EBI (approved):
Hopefully this resolves the issue with EVA-2655 too..

tcezard:
Turns out it was not necessary.

PR #351, opened 2022-07-22
EVA-2930 - Deployment for deprecation module and change study parameter name

sundarvenkata-EBI:


.gitlab-ci.yml
```diff
@@ -5,14 +5,16 @@ stages:
 
 variables:
     WS_ARTIFACT_PATH: eva-accession-ws/target/eva-accession-ws-$ENVIRONMENT_NAME.war
-    CLUSTERING_ARTIFACT_PATH: eva-accession-clustering/target/eva-accession-clustering-*.jar
+    CLUSTERING_ARTIFACT_PATH: eva-accession-clustering/target/eva-accession-clustering-*exec.jar
```
tcezard, line 8: Why are some of our jars annotated with `exec` ?

tcezard (approved):
I think it looks good although I'm a bit confused as to what is module (my confusion not the code). 
Could we use this PR to remove the `noah` deployment ?

.gitlab-ci.yml
sundarvenkata-EBI: When a module (in this case, the eva-accession-clustering module) serves as both a dependency for another module (in this case, the eva-accession-deprecate module) and also an executable (for the clustering pipeline), it has to have two deployed JARs: one that can make it behave as a "library" to be used as a dependency and another (with the exec suffix) that can be an "executable" (the one that will be deployed).

sundarvenkata-EBI:
Loosely defined, modules are units of code that are defined via a POM (project object model).  eva-accession-deprecate module comprises all the code under the directory named "eva-accession-deprecate".

sundarvenkata-EBI:
> Could we use this PR to remove the noah deployment ?

Sure. I have removed it now.

nitin-ebi (approved):

PR #356, opened 2022-08-19
EVA-2957 - Choose SS remapped from the source assembly for backpropagation

sundarvenkata-EBI:


eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/TargetSSReaderForSplitOrMergedBackPropRS.java
```diff
@@ -120,7 +125,7 @@ public List<SubmittedVariantEntity> read() {
                         .stream().map(result -> new SubmittedVariantEntity(result.getAccession(), result.getHash(),
                                                                            result.getData(), result.getVersion()))
                         .collect(Collectors.toList());
-        return targetSSRecordsInOriginalAssembly.size() == 0 ? null: targetSSRecordsInOriginalAssembly;
+        return targetSSRecordsInOriginalAssembly;
```
sundarvenkata-EBI, line 128: This was done because sometimes a RS batch we are reading might not have a corresponding SS **_in the remapped assembly_**. In that case, we will have to wait for the next batch of RS instead of returning null which will make Spring terminate the job altogether. We only have to return null and terminate the job when we have read all the RS batches - this is now made explicit a few lines above.

nitin-ebi (approved):

apriltuesday (approved):

PR #361, opened 2022-09-13
EVA-2962 - New clustered variants QC job

apriltuesday:


eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/qc/MissingCveReporter.java
```diff
@@ -0,0 +1,77 @@
...
+            if (!Objects.equals(dbRs, expectedRs)) {
+                logger.error("Hash {} has rs{} in db, expected rs{}", hash, dbRs, expectedRs);
+            }
+        }
+    }
+
+    private Map<String, ClusteredVariantEntity> findClusteredVariantsInDb(Map<String, Long> hashToRs) {
+        Query query = query(where(ID_FIELD).in(hashToRs.keySet()));
+        List<ClusteredVariantEntity> evaResults = mongoTemplate.find(query, ClusteredVariantEntity.class);
+        List<DbsnpClusteredVariantEntity> dbsnpResults = mongoTemplate.find(query, DbsnpClusteredVariantEntity.class);
```
sundarvenkata-EBI, line 72: We might have to discuss during the stand-up if we would also want to look for possibly merged RS in the clustered operations classes.

tcezard (approved):

eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/qc/MissingCveReporter.java
sundarvenkata-EBI: Based on stand-up discussions, we arrived at a consensus: Currently we are looking to find newly clustered variants at a given point in time (right after the cluster_unclustered_variants step) and this validation accomplishes that. If this code is run again after the cluster_clustered_variants step the resulting report will not have the merged RS and that is still valid.

sundarvenkata-EBI (approved):

PR #374, opened 2022-11-14
EVA-3046 & EVA-2958 - Fixes to counts scripts + adding QC script

apriltuesday:


eva-accession-release-automation/gather_clustering_counts/gather_clustering_counts_from_mongo.py
```diff
@@ -170,120 +203,145 @@ def query_mongo(mongo_source, filter_criteria, metric):
...
+        assembly_accession = previous_assembly_stats[2]
+        release_folder = previous_assembly_stats[3]
+
+        # If the complete taxonomy/assembly pair is already present in the current release, do nothing
+        if (taxonomy_id, assembly_accession) in asm_tax_in_logs:
+            continue
+
+        # If the assembly is present in the current release but not with this taxonomy,
+        # copy this release's metrics for this assembly.
+        if assembly_accession in assemblies_in_logs:
```
tcezard, line 328: And here we're retrieving the updated taxonomies populated in `fill_data_for_current_release` and using them as template.

eva-accession-release-automation/gather_clustering_counts/requirements.txt
```diff
@@ -1 +1,2 @@
 ebi_eva_common_pyutils
+pytz
```
sundarvenkata-EBI, line 2: Will be able to use the builtin [zoneinfo](https://docs.python.org/3/library/zoneinfo.html) module instead?

eva-accession-release-automation/setup.py
```diff
@@ -7,6 +7,9 @@ def get_requires():
     with open(os.path.join(os.path.dirname(__file__), "run_release_in_embassy", "requirements.txt"), "rt") as req_file:
         for line in req_file:
             requires.append(line.rstrip())
+    with open(os.path.join(os.path.dirname(__file__), "gather_clustering_counts", "requirements.txt"), "rt") as req_file:
+        for line in req_file:
+            requires.append(line.rstrip())
```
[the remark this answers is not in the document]
sundarvenkata-EBI: I agree. Is there a way to inherit a common top-level requirements.txt file and only add module specific dependencies in each module similar to Java POMs?

sundarvenkata-EBI (approved):

eva-accession-release-automation/gather_clustering_counts/requirements.txt
apriltuesday: Yes, we're on 3.8 in the cluster which is indeed why I didn't use zoneinfo...  We can use one of the other suggestions, I'm curious what's the objection to `pytz` though?  `backports` and `dateutil` also require an additional dependency.

sundarvenkata-EBI: #optional I was just suggesting that just to have one less external dependency.

eva-accession-release-automation/setup.py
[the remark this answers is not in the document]
apriltuesday: Very true.  It should be possible to have requirements files that refer directly to other requirements files, but I'm having trouble getting it to work...

Anyway I think the path of least resistance here is have a single top-level requirements file, since the dependencies are almost the same and it corresponds with how we actually use them (shared virtualenv for everything in this directory).

sundarvenkata-EBI (approved):

PR #382, opened 2023-01-26
EVA-3124 - Log discards in remapping writer

apriltuesday:
Added a couple tests to exemplify some slightly weird deduplication behaviour in the writer.  I believe it's actually correct and only applies to a situation where there are existing duplicate SS anyway, but combined with the lack of discard logging I think it could account for the counting discrepancy in the original ticket.  (I'll also create a new ticket for QC so we can continue to monitor the situation...)

### DIFF
--- a/eva-remapping-ingest/src/test/java/uk/ac/ebi/eva/remapping/ingest/batch/io/RemappedSubmittedVariantsWriterTest.java
+++ b/eva-remapping-ingest/src/test/java/uk/ac/ebi/eva/remapping/ingest/batch/io/RemappedSubmittedVariantsWriterTest.java
@@ -16,7 +16,6 @@
  */
 package uk.ac.ebi.eva.remapping.ingest.batch.io;
 
-import com.lordofthejars.nosqlunit.annotation.UsingDataSet;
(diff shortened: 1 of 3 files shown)


eva-remapping-ingest/src/test/java/uk/ac/ebi/eva/remapping/ingest/batch/io/RemappedSubmittedVariantsWriterTest.java
```diff
@@ -56,7 +55,6 @@
 @RunWith(SpringRunner.class)
 @ContextConfiguration(classes = {BatchTestConfiguration.class})
 @TestPropertySource("classpath:ingest-remapped-variants.properties")
-@UsingDataSet(locations = {"/test-data/submittedVariantEntity.json"})
```
apriltuesday, line 59: Also changed the tests so they don't rely on this dataset, as I find it easier to follow this way...

tcezard (approved):

sundarvenkata-EBI (approved):

PR #418, opened 2023-09-11
EVA-3375  set WriteResultChecking to Exception

nitin-ebi:


sundarvenkata-EBI (approved):

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/mongoTemplate/MongoTemplateTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Before
+    public void setUp() {
+
+    }
+
+    @Test(expected = DataAccessException.class)
+    public void testMongoTemplateWriteResultChecking() {
+        AnnotationMetadata annotationMetadata = new AnnotationMetadata("vep_1", "vep_cache_1", true);
+        doThrow(new DataAccessException("Simulated exception") {})
```
tcezard, line 48: `DataAccessException` is getting thrown and caught by us. 
Is it not going through Spring ?

tcezard (approved):

eva-accession-core/src/test/java/uk/ac/ebi/eva/accession/core/mongoTemplate/MongoTemplateTest.java
sundarvenkata-EBI: Come to think of it, this test shouldn't even be there because it is tautologically true. But I guess it is only there as a placeholder if we **_eventually_** figure out a way to test the WriteResultChecking.

nitin-ebi: To check **WriteResultChecking**, ideally we will need an integration test that can somehow throw an error when doing DB operations. As Sundar said, this is a just a placeholder for the time being.

PR #419, opened 2023-09-15
EVA-3378 scientific name fixes

nitin-ebi:


sundarvenkata-EBI (approved):

nitin-ebi:
Decided not to go ahead with this approach (will use the metadata api for getting the scientific name)

PR #421, opened 2023-10-16
EVA-3425 making sure reference and alternate are always upper case

nitin-ebi:


sundarvenkata-EBI (commented):
Looks good but we might need to write explicit tests given that we are modifying something very fundamental...

tcezard (approved):

PR #430, opened 2024-02-22
EVA-3434 Remove locus equality condition from DeprecatedVariantMongoReader

nitin-ebi:


eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/batch/io/deprecated/DeprecatedVariantMongoReader.java
```diff
@@ -107,18 +107,16 @@ protected List<Variant> getVariants(Document deprecatedVariant) {
...
             Collection<Document> inactiveEntitySubmittedVariant = (Collection<Document>) submittedVariantOperation
                     .get("inactiveObjects");
             Document submittedVariant = inactiveEntitySubmittedVariant.iterator().next();
-            long submittedVariantStart = submittedVariant.getLong(START_FIELD);
             String submittedVariantContig = submittedVariant.getString(CONTIG_FIELD);
             String reference = submittedVariant.getString("ref");
             String alternate = submittedVariant.getString("alt");
 
-            if (isSameLocation(contig, start, submittedVariantContig, submittedVariantStart, type)) {
+            if (contig.equals(submittedVariantContig)) {
```
nitin-ebi, line 119: removed the condition for checking the start,  but was not sure if we needed to remove the contig checking as well, so left it for now

tcezard: I think we need to also remove the contig condition. 
It does not make sense to check the contig if we don't check the start.

nitin-ebi: removed

tcezard (approved):

eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/batch/io/deprecated/DeprecatedVariantMongoReader.java
```diff
@@ -107,29 +107,19 @@ protected List<Variant> getVariants(Document deprecatedVariant) {
...
-
-        return new ArrayList<>();
+        Collection<Document> inactiveEntitySubmittedVariant = (Collection<Document>) submittedVariantOperations
+                .iterator().next().get("inactiveObjects");
+        Document submittedVariant = inactiveEntitySubmittedVariant.iterator().next();
+        String reference = submittedVariant.getString("ref");
+        String alternate = submittedVariant.getString("alt");
+
+        // Since we only need evidence that at least one submitted variant agrees
+        // with the deprecated RS in locus, we just return one variant record per RS
```
apriltuesday, line 117: We should remove the part about agreeing on locus, and perhaps move the comment to above line 110 as it helps explain why we only call `submittedVariantOperations.iterator().next()` rather than iterating through all operations.

(This is all a bit of a faff since AFAICT we just [output the rsID](https://github.com/EBIvariation/eva-accession/blob/4bbf2066781c756caca3c9ed206789c427f60ba3/eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/batch/io/deprecated/DeprecatedVariantAccessionWriter.java#L69) anyway and the whole aggregation pipeline is to confirm that the rsID was ever present for this taxonomy to begin with... but that's a separate discussion...)

nitin-ebi: done

PR #435, opened 2024-03-28
EVA-3539  - shutdown Accession Generator and release blocks once accessioning is done

nitin-ebi:


tcezard:
It think this looks good but I could not see where you would set the application_id in place of the [property-provided instance id](https://github.com/EBIvariation/eva-accession/blob/6006d9234704f923befc27b9ddfbbb0a908447b5/eva-accession-core/src/main/java/uk/ac/ebi/eva/accession/core/configuration/nonhuman/SubmittedVariantAccessioningConfiguration.java#L124). Would you do that in a separate PR ?

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/runner/RestartFailedJobTest.java
```diff
@@ -0,0 +1,281 @@
...
+    private void remediateTempVcfError() throws Exception {
+        writeToTempVCFFile(originalVcfContent);
+    }
+
+    private void useOriginalVcfFile() throws Exception {
+        inputParameters.setVcf(originalVcfInputFilePath);
+        vcfReader.setResource(FileUtils.getResource(new File(originalVcfInputFilePath)));
+    }
+
+    private void useTempVcfFile() throws Exception {
```
apriltuesday, line 229: We should be reusing these methods from `EvaAccessionJobLauncherCommandLineRunnerTest`, and perhaps add some documentation in the code as to why we had to separate these particular tests from the main class.

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/test/RecoverTestAccessioningConfiguration.java
```diff
@@ -56,10 +56,12 @@ public SubmittedVariantMonotonicAccessioningService submittedVariantMonotonicAcc
             @Autowired @Qualifier("testSubmittedVariantAccessionGeneratorRecover")
             MonotonicAccessionGenerator<ISubmittedVariant> accessionGenerator,
             @Autowired SubmittedVariantAccessioningDatabaseService databaseService) {
-        return new SubmittedVariantMonotonicAccessioningService(accessionGenerator,
+        SubmittedVariantMonotonicAccessioningService sveMonotonicAccService = new SubmittedVariantMonotonicAccessioningService(accessionGenerator,
                 databaseService,
                 new SubmittedVariantSummaryFunction(),
                 new SHA1HashingFunction());
+        sveMonotonicAccService.shutDownAccessioning();
```
apriltuesday, line 63: I still don't really understand why some tests need the accessioning to be actually shut down, and others need the shut down to be mocked as in `RestartFailedJobTest`.  I feel like we should make sure this is relatively clear in the code, so for example if someone in the future needs to add a new test, they know which class to put it in.

Even better if we could do something to simplify - e.g. does it help to move `recoverState` out of the constructor in the accession generator? Perhaps could also help with setting the application ID as Tim mentioned...

apriltuesday (commented):
I have some questions/confusion around the tests, but otherwise this looks good.

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/test/RecoverTestAccessioningConfiguration.java
nitin-ebi: To answer your question, Yes it would help in certain cases, if we move the recover state out of the constructor. Some tests will be simplified like the one for testing the recovery state.
We can move recovery out of constructor, but then we need to make sure when any accessioning generator is created, the first step it does is to run the recovery (constructor is the best way to make sure it does that) or maybe we can remove the recovery state entirely from accessioning generator. We can discuss on that when we discuss the issue of putting application instance id.

1. **Shutting down service in recover tests**
These beans are created just to insert data in DB for testing recovery and we want these to release the blocks as we will be trying to recover them, hence shutting them down. As the recovery runs in constructor of the accessioning generator, we need to put the data in the DB before accessioning generator bean is created and run recovery as part of its creation. We can't put the data as part of the test's setup method, because it will be too late and all the beans would have been created by then and recovery state would have already ran, hence the workaround of creating these dummy beans to insert data.
    
3. **Mocking(Spying) the service in RestartFailedJobTest.**
In order to test the restarting of a failed job, we need to run a job and fail it and then subsequently need to run same job again by forcing it to restart. In real world, both runs of the job will be using different instances of the services as the entire stack of beans will be recreated for the restart job but that is not the case here, where the same set of beans are used for different runs. Which means we are trying to use an accession generator which is already shut down.Could not find an easier way to insert different beans for different runs hence the workaround of not shutting down the existing generator and reuse it.

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/runner/runnerUtil.java
```diff
@@ -0,0 +1,84 @@
...
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.zip.GZIPInputStream;
+import java.util.zip.GZIPOutputStream;
+
+public class runnerUtil {
```
apriltuesday, line 20:
```suggestion
public class RunnerUtil {
```

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/test/RecoverTestAccessioningConfiguration.java
apriltuesday: Thanks for taking the time to explain it, I think the part I wasn't getting is that these beans are mostly used to prepare the data for testing. It makes sense now and I agree we can discuss pros/cons of moving recoverState later.

eva-accession-pipeline/src/test/java/uk/ac/ebi/eva/accession/pipeline/runner/runnerUtil.java
nitin-ebi: corrected

PR #454, opened 2024-06-19
EVA-3589: Set WriteConcern and ReadPreference for MongoTemplate

apriltuesday:
Do this explicitly as apparently constructing it from the factory is not enough!


apriltuesday:
I can't for the life of me figure out why the tests aren't passing, it looks like it's not mocking count-stats correctly and I'm not sure why (or why it would be triggered by the mongo template change). If one of you can figure it out please let me know!

nitin-ebi (approved):

tcezard (approved):

apriltuesday:
Thank you for fixing @nitin-ebi !

PR #459, opened 2024-07-17
Check before creating by_assembly directories

tcezard:
Only create a symlink in the by assembly directory if the corresponding by_species directory exists

### DIFF
--- a/eva-accession-release-automation/publish_release_to_ftp/publish_release_files_to_ftp.py
+++ b/eva-accession-release-automation/publish_release_to_ftp/publish_release_files_to_ftp.py
@@ -129,13 +129,16 @@ def create_symlink_to_species_folder_from_assembly_folder(current_release_assemb
     public_release_species_assembly_folder = os.path.join(release_properties.public_ftp_current_release_folder,
                                                           by_species_folder_name, species_release_folder_name,
                                                           assembly_accession)
-    run_command_with_output(f"""Creating symlink from assembly folder {public_release_assembly_species_folder} to
-                            species folder {public_release_species_assembly_folder}""",
-                            'bash -c "cd {0} && ln -sfT {1} {2}"'.format(
-                                public_release_assembly_folder,
-                                os.path.relpath(public_release_species_assembly_folder,
-                                public_release_assembly_folder),
-                                public_release_assembly_species_folder))
+    if os.path.isdir(public_release_assembly_species_folder):
+        run_command_with_output(f"""Creating symlink from assembly folder {public_release_assembly_species_folder} to
+                                species folder {public_release_species_assembly_folder}""",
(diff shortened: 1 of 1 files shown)


eva-accession-release-automation/publish_release_to_ftp/publish_release_files_to_ftp.py
```diff
@@ -198,9 +201,10 @@ def publish_assembly_release_files_to_ftp(current_release_assembly_info, release
             )
         )
 
-    # Create a link from assembly folder to species_folder ex: by_assembly/GCA_000005005.5/zea_mays to by_species/zea_mays/GCA_000005005.5
-    create_symlink_to_species_folder_from_assembly_folder(current_release_assembly_info, release_properties,
-                                                          public_release_assembly_folder)
+    if current_release_assembly_info["num_rs_to_release"] > 0:
```
apriltuesday, line 204: Debatable whether this is even a good idea, but elsewhere we check both `num_rs_to_release` and `should_be_released`... should we do the same here to be consistent?

eva-accession-release-automation/publish_release_to_ftp/publish_release_files_to_ftp.py
```diff
@@ -129,13 +129,16 @@ def create_symlink_to_species_folder_from_assembly_folder(current_release_assemb
...
+    if os.path.isdir(public_release_assembly_species_folder):
+        run_command_with_output(f"""Creating symlink from assembly folder {public_release_assembly_species_folder} to
+                                species folder {public_release_species_assembly_folder}""",
+                                'bash -c "cd {0} && ln -sfT {1} {2}"'.format(
+                                    public_release_assembly_folder,
+                                    os.path.relpath(public_release_species_assembly_folder,
+                                    public_release_assembly_folder),
+                                    public_release_assembly_species_folder))
+    else:
+        raise Exception(f'The species folder {public_release_assembly_species_folder} were linking to does not exist')
```
apriltuesday, line 141:
```suggestion
        raise Exception(f'The species folder {public_release_assembly_species_folder} we're linking to does not exist')
```

apriltuesday (approved):

nitin-ebi (approved):

eva-accession-release-automation/publish_release_to_ftp/publish_release_files_to_ftp.py
tcezard: I think you're right that this is inconsistent with what is done elsewhere.
The reason I've only used the `num_rs_to_release` is because originally the `should_be_released` flag really meant that the release pipeline should be run not that the data should be copied and in previous version of the release automation `should_be_released=false` and  `num_rs_to_release>0` was used to denote release that had not changed since the previous release and could be hard-linked.

PR #469, opened 2025-02-20
EVA-3629: Updates to process remapped variants writer

apriltuesday:
* Don't include unchanged operations in list of operations to update
* Change fetching of existing operations from aggregate to find

### DIFF
--- a/eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
+++ b/eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
@@ -21,8 +21,6 @@ import org.springframework.batch.core.JobExecution;
 import org.springframework.batch.item.ItemWriter;
 import org.springframework.data.mongodb.core.BulkOperations;
 import org.springframework.data.mongodb.core.MongoTemplate;
-import org.springframework.data.mongodb.core.aggregation.Aggregation;
-import org.springframework.data.mongodb.core.aggregation.AggregationOptions;
 import org.springframework.data.mongodb.core.query.Query;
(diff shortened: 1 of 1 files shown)


eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java
```diff
@@ -144,18 +142,14 @@ public ClusteringWriter(MongoTemplate mongoTemplate,
...
 
         try (
-                CloseableIterator<SubmittedVariantOperationEntity> svoes = mongoTemplate.aggregateStream(
-                        Aggregation.newAggregation(
-                                Aggregation.match(where("eventType").in(MERGE_AND_SPLIT_EVENTS)
-                                        .and("inactiveObjects").elemMatch(where("seq").is(assembly))))
-                        .withOptions(aggregationOptions), SubmittedVariantOperationEntity.class.getSimpleName(),
-                SubmittedVariantOperationEntity.class)) {
+                CloseableIterator<SubmittedVariantOperationEntity> svoes = mongoTemplate.stream(
+                        queryOperations, SubmittedVariantOperationEntity.class)) {
```
apriltuesday, line 152: This was changed from `find` to `aggregate` [here](https://github.com/EBIvariation/eva-accession/pull/388), but I can't see why it's necessary - if I understand the context correctly, that PR was more about using an iterator to alleviate memory issues, which should be achievable without using aggregations.

apriltuesday: Actually it looks like this aggregate simply did not work, and this is the origin of the duplicate keys - will write it up in the ticket.

nitin-ebi (approved):

PR #500, opened 2026-05-18
Security fixes for accession-ws

tcezard:
1. Path traversal in contig alias URL construction (ContigAliasService.java)
Replaced string concatenation with UriComponentsBuilder.pathSegment() + .queryParam() for referenceName and assemblyId request parameters. Added a guard rejecting the bare  dot-segments . and .. as contig identifiers.
2. Internal service URL disclosed in error responses (ContigAliasService.java, both controllers)
Changed the exception messages to include only the user-supplied identifier (not the URL), and replaced the catch-all message in both beacon endpoints with a fixed generic string, logging the full detail  server-side instead.
3. Open redirect via Host header injection (EvaControllerAdvice.java)
When a merged accession was requested, the 301 redirect was built from httpServletRequest.getRequestURL(), which Tomcat populates using the Host request header. Overrode handleMergeExceptions in EvaControllerAdvice to use getRequestURI() (path only, no host) and prepend the trusted eva.api.base-url configuration property instead.
4. Full git commit messages exposed via actuator (application.properties)
management.info.git.mode=full caused /actuator/info to include the complete last commit message in its public JSON response. Changed to management.info.git.mode=simple, which retains only the commit ID and timestamp.
5. Permissions: Added permissions: contents: read block in github actions

### DIFF
--- a/eva-accession-ws/src/main/resources/application.properties
+++ b/eva-accession-ws/src/main/resources/application.properties
@@ -22,7 +22,11 @@ human.mongodb.database=|eva.accession.mongo.human.database|
 contig-alias.url=|contig-alias.url|
 
 management.endpoints.web.exposure.include=info,health
-management.info.git.mode=full
+management.info.git.mode=simple
+
+# Base URL used to build safe redirect responses (must not end with a slash).
+# Set this to the canonical public URL of the API in each deployment environment.
+eva.api.base-url=|eva.api.base-url|
 
 # See https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.1-Release-Notes#bean-overriding
 spring.main.allow-bean-definition-overriding=true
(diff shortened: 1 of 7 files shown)


eva-accession-ws/src/main/resources/application.properties
```diff
@@ -22,7 +22,11 @@ human.mongodb.database=|eva.accession.mongo.human.database|
 contig-alias.url=|contig-alias.url|
 
 management.endpoints.web.exposure.include=info,health
-management.info.git.mode=full
+management.info.git.mode=simple
+
+# Base URL used to build safe redirect responses (must not end with a slash).
+# Set this to the canonical public URL of the API in each deployment environment.
+eva.api.base-url=|eva.api.base-url|
```
nitin-ebi, line 29: `eva.api.base-url `will need to be added in configuration (maven-settings) as well before merging this

nitin-ebi (approved):

apriltuesday (approved):

PR #511, opened 2026-07-29
Remove accessionning block dependency

tcezard:


eva-accession-ws/pom.xml
```diff
@@ -58,7 +58,6 @@
         <dependency>
             <groupId>org.hsqldb</groupId>
             <artifactId>hsqldb</artifactId>
-            <scope>test</scope>
```
tcezard, line 61: Needed outside the test because we now rely on an in-memory database

eva-accession-ws/src/test/java/uk/ac/ebi/eva/accession/ws/ReadOnlyConfigurationRestControllerTest.java
```diff
@@ -0,0 +1,105 @@
...
+import static org.mockito.ArgumentMatchers.argThat;
+import static org.mockito.Mockito.when;
+
+/**
+ * Unlike the other tests in this package, this test does NOT import
+ * {@code SubmittedVariantAccessioningConfiguration} or {@code ContiguousIdBlocksDataSourceConfiguration} (the
+ * real, JPA/PostgreSQL-backed core configuration). It only relies on the application's own default wiring
+ * (EvaAccessionApplication ApplicationConfiguration the ReadOnly* configurations), the same wiring used
+ * in production. Other historical tests imports accessioning service for seeding data, which means their Spring
+ * context always has a bean named "submittedVariantAccessioningService" already present.
```
apriltuesday, line 58: Shouldn't we update the main tests to use the configuration that we use in production, rather than add a new test? Otherwise we're not fully testing all the code paths in the read-only configuration that will actually be used. We can seed the data using the MongoTestDataLoader instead of using the services.

eva-accession-ws/src/test/java/uk/ac/ebi/eva/accession/ws/ReadOnlyConfigurationRestControllerTest.java
```diff
@@ -0,0 +1,105 @@
...
+    @Test
+    public void getSubmittedVariantThroughReadOnlyConfiguration() {
+        when(contigAliasService.translateContigToInsdc(anyString(), anyString(),
+                argThat(new NoContigTranslationArgumentMatcher()))).thenCallRealMethod();
+        when(contigAliasService.getSubmittedVariantsWithTranslatedContig(any(), any())).thenCallRealMethod();
+
+        new MongoTestDataLoader(mongoTemplate, resourceLoader).loadAll("/test-data/readOnlySubmittedVariant.json");
+
+        ResponseEntity<List<AccessionResponseDTO<SubmittedVariant, ISubmittedVariant, String, Long>>> response =
+                testRestTemplate.exchange("/v1/submitted-variants/42", HttpMethod.GET, null,
```
apriltuesday, line 91: I guess this isn't the main point, but how does this actually work, given that [readOnlySubmittedVariant.json](https://github.com/EBIvariation/eva-accession/pull/511/changes#diff-bf741d9a2277930f340c1dbed79c91d58c2375bbfe0dd8c96a30f37428e4d3cc) contains two distinct submitted variants with accession 42?

eva-accession-ws/src/main/java/uk/ac/ebi/eva/accession/ws/configuration/ReadOnlyContiguousIdBlockServiceConfiguration.java
```diff
@@ -0,0 +1,114 @@
...
+ * actually used. That injection needs an EntityManagerFactory bean to exist in the context; without one, the
+ * context fails to start even though nothing on eva-accession-ws's read paths would ever touch it. So this
+ * configuration provides its own EntityManagerFactory backed by an embedded, in-process HSQLDB instance instead
+ * of the original's PostgreSQL DataSource.
+ * The method transactionManager serves a similar purpose since ContiguousIdBlockService contains the Transactional annotation
+ *
+ * {@code @ConditionalOnMissingBean} makes this back off whenever a real, JPA-backed ContiguousIdBlockService is
+ * already present in the context (e.g. tests that explicitly import the write-capable core configuration to seed
+ * data), so it never shadows a functional bean with this read-only one.
+ */
```
apriltuesday, line 62: Do you think it would be substantially more complicated to actually change the webservice to be more explicitly read-only? For example, we shouldn’t need a block service or accession generator at all, let alone one backed by an in-memory database.

If I understand correctly (which I probably don’t, but that never stopped me!), this would entail:
- new accessioning services that only provide `get` methods and don’t require an accession generator
- some bean that provides the SS & RS bounds from the properties `accessioning.monotonic.ss.blockStartValue=5000000000` and `accessioning.monotonic.rs.blockStartValue=3000000000`
- configuration classes that only provide these new read-only beans

Right now I think such an approach would be my preference. But very likely I am missing some complications, so happy to discuss further.

PR #513, opened 2026-08-07
EVA-4242: Make accession-ws explicitly read-only

apriltuesday:
This PR replaces accessioning services with read-only equivalents, which do not have the ability to generate accessions and do not require access to PostgreSQL for the contiguous ID block service.

An alternate approach is in #511, which modifies only configuration so the block service uses an in-memory database instead. The approach in this PR requires changes to the actual service but makes the read-only functionality more explicit.

Tested using locally using docker compose as well as Kubernetes (minikube).


nitin-ebi (approved):

eva-accession-ws/src/main/java/uk/ac/ebi/eva/accession/ws/service/ReadOnlyClusteredVariantService.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Read-only version of {@link uk.ac.ebi.eva.accession.core.service.nonhuman.ClusteredVariantAccessioningService}
+ */
+public class ReadOnlyClusteredVariantService implements AccessioningService<IClusteredVariant, String, Long> {
+
+    private static Logger logger = LoggerFactory.getLogger(ClusteredVariantAccessioningService.class);
```
tcezard:
```suggestion
    private static Logger logger = LoggerFactory.getLogger(ReadOnlyClusteredVariantService.class);
```

eva-accession-ws/src/main/java/uk/ac/ebi/eva/accession/ws/service/ReadOnlySubmittedVariantService.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+
+/**
+ * Read-only version of {@link uk.ac.ebi.eva.accession.core.service.nonhuman.SubmittedVariantAccessioningService}
+ */
+public class ReadOnlySubmittedVariantService implements AccessioningService<ISubmittedVariant, String, Long> {
+
+    private static Logger logger = LoggerFactory.getLogger(SubmittedVariantAccessioningService.class);
```
tcezard:
```suggestion
    private static Logger logger = LoggerFactory.getLogger(ReadOnlySubmittedVariantService.class);
```

tcezard (approved):

